PR #2, opened 2021-04-08
Correct kpng api subcommand

rikatz:
subcommands are conflicting, and I spent like 3 hours to figure out why :P


rikatz:
do not merge yet, forgot another part here :/

rikatz:
will fix other things here before :D

PR #4, opened 2021-04-18
Develop yaml, WIP, just a beachhead to start iterating, runs kpng in …

jayunit100:
…a pod but needs svc accounts, adding later


jayunit100:
not needed, using dock/ds.yaml

PR #10, opened 2021-04-29
fix the kpng kube-proxy connectivity by mounting the kubelet host

jayunit100:
iterating, this gets us over the apiserver hump....


mcluseau:
approving but it seem to be more and more KinD specific ;)

/lgtm
/approve

PR #45, opened 2021-07-15
Atomic write/delete changes for ipvs

hanamantagoudvk:
Design principles:
------------------
a) virtual-service IPs and real-server IPs are stored in a separate btree node.
b) rendering of write/delete ipvs cmds is done synchronously to make
sure that no back-to-back changes overstep on each other.

### DIFF
--- a/backends/ipvs/buffer.go
+++ b/backends/ipvs/buffer.go
@@ -1,0 +1,209 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


mcluseau:
I think that, as is, the diff tree is completely useless to detect changes. The idea of this tree in nft was to write buffers and use hashes to know what has changed. Here, `Write*` are just assigning a new value without touching the hash. It's probably easier to use the `filterreset` to accumulate the changes and apply them on sync (or when your batch is getting big). What about doing a hands-on session tomorrow at the meeting, where I'd try to implement this backend?

hanamantagoudvk:
Closing this PR because mikael added code changes with filterrest+decoder

PR #51, opened 2021-08-02
cleanup and compile fixes, also removing nodetracker stuff since kpng…

jayunit100:
… does it for us


jayunit100:
minor compilation and cleanup fixes on the iptables stuff

PR #64, opened 2021-09-03
Include lb and session affinity related parameters to kpng controller

VivekThrivikraman-est:


pkg/api/localnetv1/services.proto
```diff
@@ -82,6 +82,15 @@ message Service {
     repeated PortMapping Ports = 6;
 
     bool ExternalTrafficToLocal = 7;
+
+    string SessionAffinity = 10;
```
mcluseau, line 86: remove from here and move to a `SessionAffinity` message...

pkg/api/localnetv1/services.proto
```diff
@@ -82,6 +82,15 @@ message Service {
...
+    SessionAffinityConfig SessionAffinityConfig = 11;
```
mcluseau, line 87: ... rename to SessionAffinity...

pkg/api/localnetv1/services.proto
```diff
@@ -82,6 +82,15 @@ message Service {
...
+    IPSet LoadBalancerSourceRanges = 12;
+    IPSet LoadBalancerIngressIPs = 13;
+}
+
+message SessionAffinityConfig {
```
mcluseau, line 92: rename to SessionAffinity and add `string Kind = 1` that will receive the k8s `SessionAffinity` value

mcluseau (changes requested):

qabpea:
Session affinity is potentially tricky. There are situations where it does not scale very well, at L4 this typically occurs when there is L4 non transparent proxy or NAT in the way. I can see why you would do this with external IPs, but on the inside  of the cluster there is a potential problem when SNATing a node. The affinity must be set before the SNAT occur, if not then all sessions from one node will be directed towards one POD. This is currently not a problem, but it can/will become a problem if the order of LB and SNAT is not done in the correct order (or even better, never use SNAT for cluster internal traffic)

jayunit100:
fixes #93 #64

mcluseau:
plz test with latest master

jayunit100:
all done now by mikael

PR #74, opened 2021-10-12
Add documentation of what the service proxy does

danwinship:
There was a bug in ovn-kubernetes with respect to external IP handling, and someone was asking me what the expected behavior was, and I was like "well let me re-read kube-proxy code" and he was like "uh, that can't be the definition of the expected behavior" and I was like "well, yeah, actually it is".

We should try to do better than that. Here's an attempt. (I basically just read through the iptables proxy's `syncProxyRules` and wrote down everything it does, with some clarifications added based on some "mistakes" ovn-kubernetes had made.)

I'm submitting this to the kpng repo since this seems like the place, in the future, where people who are implementing their own proxies would start, but maybe it doesn't really belong here. I'm going to bring this up at the SIG meeting this week...

### DIFF
--- a/doc/service-proxy.md
+++ b/doc/service-proxy.md
@@ -1,0 +1,386 @@
+# So You Want To Write A Service Proxy...
+
+Kube-proxy is the default implementation of service proxying in
+Kubernetes, but it is possible for a network plugin to implement their
+own proxy that is better or faster or that integrates better with the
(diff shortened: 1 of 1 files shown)


mcluseau:
thanks, we take as much documentation as we can :-)

/approve
/lgtm

Issue #50, opened 2021-08-01
GH Actions tests (but lets keep mob testing also)


jayunit100:
imo testing every friday  as a group is a good routine --- 
 but we should have some bot tests just to make @aojea happy :) also

UPDATE

To get started, go into every file w a go.mod and run 
```
GOOS=linux go build -o kpng ./kpng
```
in GH Actions 
@FriedrichWilken

jayunit100:
as a starting point

- backend_torture.go
- https://github.com/K8sbykeshed/k8s-service-lb-validator/ 
- sonobuoy `sig-network` 
- sonobuoy `Conformance`

should be good enough but we can skip some tests (dualstack) and so onthat arent implemented yet

aojea:
> but we should have some bot tests just to make @aojea happy :) also

users will be happier than me ...

mcluseau:
I've set a basic GitHub action to run the go tests, that a first thing ;)

jayunit100:
@rajaskakodar

hanlins:
Testing is important, let me see what I can do...
/assign

jayunit100:
thanks Hanlins we missed you !!!!!!

FriedrichWilken:
https://github.com/kubernetes-sigs/kpng/pull/85

jayunit100:
this is done now thanks to friedrich and neha !

PR #85, opened 2021-10-29
[wip] github actions to build the backend packages

FriedrichWilken:
This addresses [this issue](https://github.com/kubernetes-sigs/kpng/issues/50). It's an implementation of some simple GitHub actions that try to `go build` the backend packages. `go build` on packages discards any binaries immediately after, only giving feedback IF there was an issue while trying to build.


FriedrichWilken:
TODO add win and macos
```    runs-on: ubuntu-latest ```

**Comment by mcluseau** [comment]
backends will mostly be OS-specific, so I think the matrix approach will not work for this test. But building `kpng`  should, as backends will simply be added in their os-specific `storecmd_xxx.go`.

**Comment by FriedrichWilken** [comment]
> backends will mostly be OS-specific, so I think the matrix approach will not work for this test. But building `kpng` should, as backends will simply be added in their os-specific `storecmd_xxx.go`.

I am not sure if I get this right. Do you say: don't try to build it on all OS'. Or is just this matrix approach not working?

**Comment by FriedrichWilken** [comment]
Because it is a little confusing to find the gh action results between different commits and branches: [this is the result of the test of the last commit](https://github.com/FriedrichWilken/kpng/actions/runs/1427932363)

File: .github/workflows/backend-build.yml [hunk 1]
```diff
@@ -0,0 +1,83 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: checkout
+      uses: actions/checkout@v2
+    
+    - name: Build backends/nft
+      run: |
+        cd ./backends/nft
+        go mod vendor
+        go build
```
jayunit100, line 83: imo this is a good start !

jayunit100:
lets get these passing and then merge !

jayunit100:
for now we can ignore windows, merge and iterate ! thank you friedrich !

jayunit100:
as a FOLLOW ON can you or neha break the code out into a Makefile?

.github/workflows/backend-build.yml
mcluseau: wouldn't `go mod download` be enough? Sorry I tend to be a semantic nazi :p

jayunit100:
Ok thanks for prototypeing now we know what to do #96

FriedrichWilken:
This PR was made obsolete by @nehaLohia27 's [PR](https://github.com/kubernetes-sigs/kpng/pull/97) thus I will close it.

PR #100, opened 2021-11-09
more generic build.sh

FriedrichWilken:
I changed the logic in `build.sh` so new backend packages would automatically also be tested and if 'old' backends get removed nothing breaks. On purpose, this only changes the code that is NOT used by GH actions.

I also added the license boilerplate text.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -36,8 +36,7 @@ function build_package() {
 
 function build_all_backends() {
   cd ./backends
-  for f in $(find . -type f -name 'go.mod' | sed -r 's|/[^/]+$||' |sort |uniq)
-    do build_package $f; done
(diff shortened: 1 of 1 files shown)


hack/build.sh
```diff
@@ -23,9 +35,9 @@ function build_package() {
 }
 
 function build_all_backends() {
-  build_package backends/iptables
-  build_package backends/ipvs-as-sink
-  build_package backends/nft
+  cd ./backends
+  for f in $(find . -type f -name 'go.mod' | sed -r 's|/[^/]+$||' |sort |uniq)
```
mcluseau, line 39: I think you can do a simpler `for f in $(find -name go.mod); do build_package $(dirname $f); done` (didn't test though ;))

FriedrichWilken: Ha! You are right. :)

PR #105, opened 2021-11-11
adding conformance tests

FriedrichWilken:
First iteration of sonobuoy tests. Just run ``` ./hack/test_sonobuoy.sh ``` It should set up the whole test env and run the sonobuoy quick test. 

In future iterations, we could add more tests and run them via GH Actions.

### DIFF
--- a/hack/setup_sonobuoy.sh
+++ b/hack/setup_sonobuoy.sh
@@ -14,8 +14,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-for f in $(find -name go.mod);
-  do d=$(dirname $f); 
-    echo "downloading mods in $d"; 
-    ( cd $d && go mod download ); 
-  done
+function is_kind_available() {
+  if ! command -v kind &> /dev/null
+  then
+    echo "kind could not be found"
+    exit
+  fi
+}
+
+function get_sonobuoy() {
+  mkdir sonobuoy
+  wget https://github.com/vmware-tanzu/sonobuoy/releases/download/v0.55.0/sonobuoy_0.55.0_linux_amd64.tar.gz -O ./sonobuoy.tar.gz
+  tar -xf sonobuoy.tar.gz -C ./sonobuoy
+  rm sonobuoy.tar.gz
+}
+
+is_kind_available
+
+# cd to dir of script
+cd "${0%/*}"
+
+get_sonobuoy
+
+# create a kind cluster with kpng instead of kubeproxy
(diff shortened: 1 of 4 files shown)


hack/setup_sonobuoy.sh
```diff
@@ -0,0 +1,53 @@
...
+  then
+    echo "kind could not be found"
+    exit
+  fi
+}
+
+function setup_sonobuoy() {
+  if [ -f ./sonobuoy/sonobuoy ]; then
+    echo "Found Sonobuoy, skipping setup."
+  else
```
knabben, line 28:
``` if ... then
      echo "Found Sonobuoy, skipping setup."
      exit 1
   fi
```

hack/setup_sonobuoy.sh
```diff
@@ -0,0 +1,53 @@
...
+    architecture=$(dpkg --print-architecture)
+    url="https://github.com/vmware-tanzu/sonobuoy/releases/download/v0.55.0/sonobuoy_0.55.0_"$os"_"$architecture".tar.gz"
+    wget $url -O ./sonobuoy.tar.gz
+    tar -xf sonobuoy.tar.gz -C ./sonobuoy
+    rm sonobuoy.tar.gz
+  fi
+}
+
+function setup_kind() {
+  if kind get clusters | grep -q kpng-proxy; then
```
knabben, line 40: can be a good issue for kind the creation of `kind get cluster --name kpng-proxy` attribute

hack/setup_sonobuoy.sh
jayunit100: will do this in a follow on , lets merge and iterate to get this in before the mtng today

FriedrichWilken: okay, @jayunit100. @nehaLohia27 could you please fix this in your further work on sonobuoy?

hack/setup_sonobuoy.sh
FriedrichWilken: @nehaLohia27 again, could you please fix this in your further work on sonobuoy? 🙂

FriedrichWilken: good point

hack/setup_sonobuoy.sh
jayunit100: yup merged making pr now

jayunit100: ah wait i did e 1 and that fails for other reasons bc i have sonobuoy already

PR #114, opened 2021-11-16
[wip] adding antrea to kpng-local-up.sh

FriedrichWilken:
```
export CNI=antrea
./hack/kpng.local-up.sh
```

I temporarily set antrea as default. This should be set back to calico before merging. But until then, there is more work to do.


jayunit100:
reopen if necessary @FriedrichWilken 
/close

PR #115, opened 2021-11-18
Fixing service event handling

hanamantagoudvk:
When port is deleted from service , the service event handling code used to crash due to wrong index check.


mcluseau:
oops :-)

/lgtm
/approve

PR #118, opened 2021-11-19
run e2e test with kind

aojea:
This adds presubmit jobs to run a cluster with the following matrix:
- cluster ipv4, ipv6 dual
- kpng iptables and nft

It takes less than 30 mins and it gives good coverage at least to check that it has feature parity with kube-proxy

There are test failing that need to be addresses,  an example run in my fork

https://github.com/aojea/kpng/actions/runs/1481423799

The good thing is that anybody copy this file to their fork and run the actions there

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,207 @@
+name: e2e
+
+on:
+  push:
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+
+env:
+  GO_VERSION: "1.17.3"
+  K8S_VERSION: "v1.22.2"
+  KIND_CLUSTER_NAME: "kpng"
+
+jobs:
+  build:
+    name: build
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: ${{ env.GO_VERSION }}
+      id: go
+
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Build
+      run: |
+        docker build -t kpng:test -f Dockerfile .
+        mkdir _output
+        docker save kpng:test > _output/kpng-image.tar
+
+    - uses: actions/upload-artifact@v2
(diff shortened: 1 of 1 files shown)


aojea:
/hold

I think you should fix the failing tests before merging

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,207 @@
...
+      IP_FAMILY: ${{ matrix.ipFamily }}
+      BACKEND: ${{ matrix.backend }}
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2
+
+    - name: Enable ipv4 and ipv6 forwarding
+      run: |
+        sudo sysctl -w net.ipv6.conf.all.forwarding=1
+        sudo sysctl -w net.ipv4.ip_forward=1
```
FriedrichWilken, line 61: @jayunit100 how we handle this with our kind cluster for local tests? 
I don't want to mess around with people's docker settings on their actual host machines: 
https://docs.docker.com/config/daemon/ipv6/ 
Also, we would need to activate with ipv6 on the actual host machines.
https://kind.sigs.k8s.io/docs/user/configuration/#networking

My idea: we just check if the host + docker are supporting ipv6 with something like this 
```
cat /sys/module/ipv6/parameters/disable
```
and this for ipv4 ipforward
```
sysctl net.ipv4.ip_forward
```
Then we give a warning out and pass the affected tests. In that case, we should address it in the README.

aojea: I think you can omit this if you want to distribute this, just document it or add a sanity check for failing, this is specific of this job

jayunit100: agree, thanks for noticing this friedrich - our style of dev is heavily based on the local dev stuff so lets add it in a way that we know it will still work locally - and then automate running it in GH Actions

Great catch

jayunit100:
@FriedrichWilken ~ is decomposing this into shell scripts that we can call locally as needed 

thanks @aojea !

PR #121, opened 2021-11-24
implementation of local e2e-tests

FriedrichWilken:
Based on [Antonio's PR](https://github.com/kubernetes-sigs/kpng/pull/118), this PR is introducing local e2e tests. This is also supposed to provide a base to build CI from, so that local tests and CI tests are equivalent.

To tests this, run 

```
./hack/test_e2e.sh
```


README.md
```diff
@@ -73,3 +73,6 @@ Participation in the Kubernetes community is governed by the [Kubernetes Code of
 
 [owners]: https://git.k8s.io/community/contributors/guide/owners.md
 [Creative Commons 4.0]: https://git.k8s.io/website/LICENSE
+
+
+TODO add some lines about what to find ./hack
```
knabben, line 78: a `tree` command output and the description of each file could be helpful on this PR

hack/test.sh
```diff
@@ -0,0 +1,15 @@
+E2E_K8S_VERSION="v1.22.2"
+IP_FAMILY="ipv4"
+cat <<EOF | kind create cluster \
+    --name lol           \
```
knabben, line 4: lets keep a better naming :) if we have more cluster better to not call it  `lol2` or `rofl`

hack/test_e2e.sh
```diff
@@ -0,0 +1,277 @@
...
+    echo "  Building the cluster '${E2E_CLUSTER_NAME}'"
+    line
+
+    # Get rid of any old cluster with the same name.
+    if kind get clusters | grep -q ${E2E_CLUSTER_NAME}; then
+        kind delete cluster --name ${E2E_CLUSTER_NAME}
+    fi
+    # create cluster
+
+    cat <<EOF | kind create cluster \
```
knabben, line 130: can we reuse the `hack/test.sh` command here or centralize this configuration writing somewhere?

hack/test_e2e.sh
```diff
@@ -0,0 +1,277 @@
...
+    export IMAGE=kpng:test
+    export PULL=IfNotPresent
+    export BACKEND=${E2E_BACKEND}
+    envsubst <${0%/*}/kpng-deployment-ds.yaml.tmpl >${E2E_ARTIFACTS}/kpng-deployment-ds.yaml
+    echo "deploy ds"
+    kubectl create -f ${E2E_ARTIFACTS}/kpng-deployment-ds.yaml
+    echo "wait"
+    kubectl --namespace=kube-system rollout status daemonset kpng -w --timeout=3m
+
+    echo -e "Let's move on.\n"
```
knabben, line 227: what about something informing `kpng is installed.`

README.md
FriedrichWilken: I think the need to rework of the readme is bigger than this PR, so I created a new [issue](https://github.com/kubernetes-sigs/kpng/issues/124). (This is completely not me being too lazy for writing docs. :) )

FriedrichWilken:
@nehaLohia27 let's take a look at the skipped tests. Maybe we should make them configurable via args we pass to the script.

nehaLohia27:
@FriedrichWilken Yess we can take a input for the skipped flags. but for now we can keep it common for all backends. once we have all features implemented then we can add this via args.. what do you think ?

jayunit100:
is this ready to merg?

jayunit100:
@FriedrichWilken can you squash ?

FriedrichWilken:
@nehaLohia27  can you 
```
/lgtm
/approve
```
looks like this is working now

VivekThrivikraman-est:
Currently the tests are run for every PR and since we haven't covered all testcases in any of the backends, the tests would fail and we would not be able to merge any change. Can we make tests for PR's optional for now and also have an option to have builds periodically or trigger build manually or as @hanamantagoudvk suggested as a trigger from github actions?

hanamantagoudvk:
My suggestion is we can run two types of jobs/tests if its feasible with given infra we have.

a) Cron Job : To run nightly test suites which will have elaborate coverage. 
But this might be difficult to setup and might take more time.
We can explore if its possible at all.

b) Backend specific suites : These backend specific suites can be coupled with keywords.
Whenever developer raises PRs , they can use these keywords and only specific suite is run.
With this approach , it will be easy to merge the PRs.
Currently all backend suites are running , and hence blocking merger of every PR.


Option - (b) is possible  as its already being done in k8s .

PR #159, opened 2021-12-21
test_e2e: ci_mode is a variable not cmd

dougsland:
Fix the issue:
./test_e2e.sh: line 119: ci_mode: command not found

### DIFF
--- a/hack/test_e2e.sh
+++ b/hack/test_e2e.sh
@@ -116,7 +116,7 @@ function setup_ginkgo {
 }
 
 function apply_ipvx_fixes {
-    if ci_mode = true ; then
+    if "${ci_mode}" = true ; then
         sudo sysctl -w net.ipv6.conf.all.forwarding=1
         sudo sysctl -w net.ipv4.ip_forward=1
     fi


hack/test_e2e.sh
```diff
@@ -116,7 +116,7 @@ function setup_ginkgo {
 }
 
 function apply_ipvx_fixes {
-    if ci_mode = true ; then
+    if "${ci_mode}" = true ; then
```
knabben, line 119: @FriedrichWilken what's the deal with `ci_mode` here?

nehaLohia27: @knabben We use this check to know this function should run in github ci or not

anusha94: Why can't it be made common across local and CI?
I see 3 usages of this variable
1. it is installing kind, kubectl etc if ci_mode is true. We can just check if binary not present, then install.
2. ip forwarding settings - just echo saying we are setting it to true for the test (if you think this shouldn't be modified on the host, then how about something like preserving what the initial value was and then restoring it after the test?)
3. clean artifacts - I'm not quite sure what cleanup it is doing. But need not be separate for CI and local IMO.

dougsland: For item 2. I agree that we need to restore. For item 3. Not sure either, currently, it's exporting logs :)
By the way, I would deprecate this PR in favor of: https://github.com/kubernetes-sigs/kpng/pull/163

dougsland: I will just drop this patch in favor of 163

PR #161, opened 2021-12-21
test_e2e: Add container engine validation

dougsland:
Developers might not use only docker but different
containers engine like podman. This patch adds support to it.

### DIFF
--- a/hack/test_e2e.sh
+++ b/hack/test_e2e.sh
@@ -21,8 +21,9 @@ shopt -s expand_aliases
 : ${E2E_TIMEOUT_MINUTES:=100}
 
 OS=$(uname| tr '[:upper:]' '[:lower:]')
+CONTAINER_ENGINE="docker"
 
-if_error_exit() {
+function if_error_exit {
     ###########################################################################
     # Description:                                                            #
     # Validate if previous command failed and show an error msg (if provided) #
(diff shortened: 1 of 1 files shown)


hack/test_e2e.sh
```diff
@@ -40,6 +41,44 @@ if_error_exit() {
...
+    #   None                                                                  #
+    ###########################################################################
+
+    echo "* Detecting container engine..."
+    # If docker is not available, let's check if podman exists
+    ${CONTAINER_ENGINE} &> /dev/null
+    if [ "$?" != "0" ]; then
+        CONTAINER_ENGINE="podman"
+        ${CONTAINER_ENGINE} --help &> /dev/null
+        if_error_exit "Unable to detect container engine, try docker or podman"
```
anusha94, line 77: How about phrasing the message as 
`The e2e tests currently support docker and podman as the container engine. Please install either of them`?

anusha94 (commented):
lgtm.

hack/test_e2e.sh
dougsland: make sense

dougsland: sure, make sense. thanks for the feedback.

nehaLohia27 (approved):

PR #163, opened 2021-12-22
e2e: improvements

dougsland:
- e2e: use kubectl --all-namespaces
- e2e: specify image registry for kindest/node
- e2e: replace --timeout to --request-timeout
- e2e: show error if kind cannot load image
- e2e: fail if kind create cluster fails

### DIFF
--- a/hack/test_e2e.sh
+++ b/hack/test_e2e.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+# shellcheck disable=SC2181,SC2155,SC2128
 #
 # Copyright 2021 The Kubernetes Authors.
 # 
@@ -16,9 +17,9 @@
 
 shopt -s expand_aliases
 
-: ${E2E_GO_VERSION:="1.17.3"}
-: ${E2E_K8S_VERSION:="v1.22.2"} 
-: ${E2E_TIMEOUT_MINUTES:=100}
+: "${E2E_GO_VERSION:="1.17.3"}"
+: "${E2E_K8S_VERSION:="v1.22.2"}"
+: "${E2E_TIMEOUT_MINUTES:=100}"
 
 OS=$(uname| tr '[:upper:]' '[:lower:]')
 CONTAINER_ENGINE="docker"
@@ -73,7 +74,7 @@ function pass_message {
     # Arguments:                                                              #
     #   $1 - message to output                                                #
     ###########################################################################
-    if ! [ -n "$1" ]; then
+    if [ -z "${1}" ]; then
         echo "pass_message() requires a message"
         exit 1
     fi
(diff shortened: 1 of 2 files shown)


dougsland:
> @dougsland: Cannot trigger testing until a trusted user reviews the PR and leaves an
> `/ok-to-test` message.

Ok, I can't trigger /ok-to-test but I can't see how the patches are related to this issue.

hack/test_e2e.sh
```diff
@@ -23,6 +23,9 @@ shopt -s expand_aliases
 OS=$(uname| tr '[:upper:]' '[:lower:]')
 CONTAINER_ENGINE="docker"
 
+# Users can specify docker.io, quay.io registry
+KINDEST_NODE_IMAGE="docker.io/kindest/node"
```
anusha94, line 27: Nice! Should we also allow this to be a command line arg with default set to docker.io? What do you think?

hack/test_e2e.sh
```diff
@@ -282,7 +289,10 @@ function install_kpng {
     # TODO move this to ci:
     # docker load --input kpng-image.tar
     echo "loading kpng:test docker image"
-    kind load docker-image kpng:test --name ${E2E_CLUSTER_NAME}
+    CMD_KIND_LOAD_KPNG_TEST_IMAGE=("kind load docker-image kpng:test --name ${E2E_CLUSTER_NAME}")
+    ${CMD_BUILD_IMAGE}
```
anusha94, line 293: why do we need `CMD_BUILD_IMAGE` here? If I understand correctly, that's part of `docker_build` function right?
https://github.com/kubernetes-sigs/kpng/blob/8496fd72894dc6e887e8d0242f8cb6ba32572b6b/hack/test_e2e.sh#L86-L100

hack/test_e2e.sh
dougsland: I agree but in the first place I would ship as static. In the next weeks we can add such feature as well.

hack/test_e2e.sh
dougsland: Good catch, probably a buffer in my copy/past. Thanks!

jayunit100:
should we merge it ?

dougsland:
> should we merge it ?

Yes, If you folks agree of course :)

dougsland:
@anusha94 @jayunit100 I believe I am fine with the test_e2e.sh suggestions now. If the patch got approved and merged I am going to switch context and start looking into the tests with Petr as discussed in the last meeting. Happy Holidays!

dougsland:
The failures are not related to the patch but there are tests indeed not ready/green yet.

anusha94 (approved):
/lgtm

dougsland:
@anusha94 could you please help force merge as the jobs are failing not related to the patch? Thanks

anusha94:
> @anusha94 could you please help force merge as the jobs are failing not related to the patch? Thanks

@dougsland I don't have merge rights. Maybe @jayunit100 or @knabben can merge this?

Issue #156, opened 2021-12-17
Build a unit test that serializes a sessionAffinity Service


jayunit100:
Lets make pkg/proxystore/proxystore_test.go

Right now, this function fails:
```
// Services funcs

func (tx *Tx) SetService(s *localnetv1.Service, topologyKeys []string) {
        si := &localnetv1.ServiceInfo{
                Service:      s,
                TopologyKeys: topologyKeys,
                Hash: tx.s.hashOf(&localnetv1.ServiceInfo{
                        Service:      s,
                        TopologyKeys: topologyKeys,
   
             }),
        }

        tx.set(&KV{
                Set:       Services,
                Namespace: s.Namespace,
                Name:      s.Name,
                Value:     si,
                Service:   si,
        })
}
```

The reason if fails is because:

```
        github.com/gogo/protobuf@v1.3.2/proto/table_marshal.go:2990 +0x385
sigs.k8s.io/kpng/server/pkg/proxystore.(*Store).hashOf(0xc000285810, {0x1a8f0b0, 0xc0000a38b0})
        sigs.k8s.io/kpng/server@v0.0.0-00010101000000-000000000000/pkg/proxystore/proxystore.go:69 +0x33
sigs.k8s.io/kpng/server/pkg/proxystore.(*Tx).SetService(0xc000499140, 0xc0000246c0, {0x0, 0x0, 0x0})
        sigs.k8s.io/kpng/server@v0.0.0-00010101000000-000000000000/pkg/proxystore/proxystore.go:254 +0xad
sigs.k8s.io/kpng/server/jobs/kube2store.(*serviceEventHandler).onChange.func1(0x0)
        sigs.k8s.io/kpng/server@v0.0.0-00010101000000-000000000000/jobs/kube2store/service-event-handler.go:143 +0x1ae
sigs.k8s.io/kpng/server/pkg/proxystore.(*Store).Update(0xc000285810, 0xc0002c1ce0)
        sigs.k8s.io/kpng/server@v0.0.0-00010101000000-000000000000/pkg/proxystore/proxystore.go:89 +0xc6
```

Lets make a unit test that just serializes a service of this type: 
https://zhwt.github.io/yaml-to-go/ -> 
```
apiVersion: v1
kind: Service
metadata:
  name: session-affinity-service
  namespace: default
spec:
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  ports:
  - name: http
    nodePort: 31279
    port: 80
[...]

PR #165, opened 2021-12-24
(lets make this test fail) stub of initial unit test for sessionAffinity

jayunit100:
#156


server/pkg/proxystore/proxystore_test.go
```diff
@@ -77,3 +77,51 @@ func Example() {
...
+                Hash:
+                tx.s.hashOf(&localnetv1.ServiceInfo{
+                        Service:      s,
+                        TopologyKeys: topologyKeys,
+
+                }),
+        }
+
+	input := AutoGenerated{}
+	SetService(input, []string{})
```
jayunit100, line 125: am hoping that this will, always just fail once this compiles :)

PR #167, opened 2021-12-27
go-mod-tidy-all: add validations

dougsland:
Check if go bin exists and if it's the version
that introduced -compat for go mod tidy.


dougsland:
Extra reviewers: @anusha94 @nehaLohia27

PR #169, opened 2021-12-28
test_e2e: create -d for development mode

dougsland:
test_e2e is nice tool to create kpng enviroment.
However, it always run e2e tests (as expected) which
consumes a bunch of time. This patch adds -d to skip e2e
and allow fast env creation for developers.

### DIFF
--- a/hack/test_e2e.sh
+++ b/hack/test_e2e.sh
@@ -519,6 +519,11 @@ function main {
 
     install_kpng
 
+    if [ "${devel_mode}" = true ] ; then
+	pass_message "KPNG development env created, use kubectl in the namespace ${NAMESPACE} to manage it."
+	exit 0
+    fi
+
     run_tests
 
     if [ "${ci_mode}" = false ] ; then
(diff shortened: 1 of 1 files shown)


dougsland:
Adding @anusha94 in the loop for review as well.

hack/test_e2e.sh
```diff
@@ -519,7 +519,12 @@ function main {
 
     install_kpng
 
-    run_tests
+    if [ "${devel_mode}" = true ] ; then
+	pass_message "KPNG development env created, use kubectl in the namespace ${NAMESPACE} to manage it."
+	exit 0
+    else
```
anusha94, line 525: Is `else` block really needed? You're doing an `exit 0` anyway

anusha94 (commented):
Super useful PR. Thank you!
Lgtm apart from a minor nit.

hack/test_e2e.sh
dougsland: A good call. :) Let me fix this.

dougsland:
@anusha94 looks like your approve didn't work today. :/

anusha94 (approved):

Issue #149, opened 2021-12-10
protobuf error related to sessionAffinity in CI runs


knabben:
Daemonset pods started to crashloopback and error, with the following logs:

```
oroutine 32 [running]:                                                                                                                                                                                                                        
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x164baa0, 0x268b1a0})                                                                                                                                                                          
        k8s.io/apimachinery@v0.22.2/pkg/util/runtime/runtime.go:74 +0x7d                                                                                                                                                                       
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0xc000204280})                                                                                                                                                                     
        k8s.io/apimachinery@v0.22.2/pkg/util/runtime/runtime.go:48 +0x75                                                                                                                                                                       
panic({0x164baa0, 0x268b1a0})                                                                                                                                                                                                                  
        runtime/panic.go:1038 +0x215                                                                                                                                                                                                           
github.com/gogo/protobuf/proto.makeOneOfMarshaler.func1({0x1}, 0x1)                                                                                                                                                                            
        github.com/gogo/protobuf@v1.3.2/proto/table_marshal.go:2598 +0x7b                                                                                                                                                                      
github.com/gogo/protobuf/proto.(*marshalInfo).size(0xc000616080, {0x8c4c3d})                                                                                                                                                                   
        github.com/gogo/protobuf@v1.3.2/proto/table_marshal.go:200 +0x187                                                                                                                                                                      
github.com/gogo/protobuf/proto.makeMessageMarshaler.func1({0x1000000008b0ec0}, 0x1)                                                                                                                                                            
        github.com/gogo/protobuf@v1.3.2/proto/table_marshal.go:2430 +0x39                                                                                                                                                                      
github.com/gogo/protobuf/proto.(*marshalInfo).size(0xc000616000, {0xc000205770})                                                                                                                                                               
        github.com/gogo/protobuf@v1.3.2/proto/table_marshal.go:200 +0x187                                                                                                                                                                      
github.com/gogo/protobuf/proto.(*InternalMessageInfo).Size(0x203000, {0x1a92970, 0xc000205770})                                                                                                                                                
        github.com/gogo/protobuf@v1.3.2/proto/table_marshal.go:131 +0x4a                                                                                                                                                                       
[...]

PR #171, opened 2021-12-29
 Using golang protobuf library

knabben:
Fixes https://github.com/kubernetes-sigs/kpng/issues/149

`server/pkg/proxystore/proxystore.go` and `watchstate` are using gogo library (https://github.com/gogo/protobuf) to parse the protobuf, this version is not fully compatible with the current binary build with (https://pkg.go.dev/google.golang.org/protobuf/proto), this PR changes both files to use `google protobuf` instead.


server/pkg/server/watchstate/watchstate.go
```diff
@@ -101,16 +99,16 @@ func (w *WatchState) send(item *localnetv1.OpItem) {
 }
 
 func (w *WatchState) sendSet(set localnetv1.Set, path string, m proto.Message) {
-	w.pb.Reset()
```
jayunit100, line 104: wait , no more reset ? whys that

knabben: adding a reset at this point will clear new requests..

knabben: this reset in particular is from the buffer and not the message..
adding a `proto.Reset(m)` at this point will clear new request so they come up as empty.

server/pkg/proxystore/proxystore.go
```diff
@@ -61,16 +59,16 @@ func New() *Store {
...
 func (s *Store) hashOf(m proto.Message) (h uint64) {
-	if err := s.pb.Marshal(m); err != nil {
+	message, err := proto.Marshal(m)
+	if err != nil {
 		panic(err) // should not happen
 	}
-	h = xxhash.Sum64(s.pb.Bytes())
-	s.pb.Reset()
+	h = xxhash.Sum64(message)
+	proto.Reset(m)
```
mcluseau, line 71: the original `Reset` was resetting the *buffer*, we don't want to reset the given message ;-) With `proto.Marshal` you have a new []byte every time, so no need to reset anything.

server/pkg/server/watchstate/watchstate.go
```diff
@@ -31,8 +31,7 @@ type WatchState struct {
 	res   localnetv1.OpSink
 	sets  []localnetv1.Set
 	diffs []*diffstore.DiffStore
-	pb    *proto.Buffer
-	Err   error
+	Err error
```
mcluseau, line 34: I think there's a minor go fmt issue here, `error` should be aligned.

mcluseau (changes requested):

PR #186, opened 2022-01-08
Default value for masqueradeAll flag for iptables

### LINKED ISSUE #162: [e2e debug] Preserve source pod IP for traffic thru service cluster IP
This is one of the failures reported by #132 

```
[Fail] [sig-network] Services [It] should preserve source pod IP for traffic thru service cluster IP [LinuxOnly] 
/workspace/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/network/util.go:133
```

anusha94:
Setting the default value to false.
A default value of true implied that all requests are snat-d which
prevented from preserving the source ip even for traffic from within the
cluster.

Fixes #162

### DIFF
--- a/backends/iptables/iptables.go
+++ b/backends/iptables/iptables.go
@@ -2,6 +2,7 @@ package iptables
 
 import (
 	"bytes"
+	"flag"
 	"fmt"
 	"net"
 	"strconv"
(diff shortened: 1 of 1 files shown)


anusha94:
cc @knabben @nehaLohia27

knabben:
/lgtm
we should start to think on command-line arguments at this point.

PR #190, opened 2022-01-09
externalIP implementation for cluster ip service

hanamantagoudvk:
ExternalIP programming in IPVS , ipset , iptable is being taken care with this implementation


backends/ipvs-as-sink/common.go
```diff
@@ -30,16 +31,56 @@ import (
...
+	AddEndPoint    Operation = 2
+	DeleteEndPoint Operation = 3
+)
+
 type endPointInfo struct {
 	endPointIP      string
 	isLocalEndPoint bool
 }
 
+func asDummyIPs(ip string, ipFamily v1.IPFamily) string {
```
qabpea, line 60: Two things, 
First why the last ip + "/32" which other value can ipFamily have? Any value other than v4 or v6 should be an error.
Secondly, why the need to prepend a /32 or /128 scope at all? Do you just want to get "vip" address that you can hang on a loopback or use in route.
Given that the name of the function, this is just for my curiosity. No need really to change anything

qabpea (approved):
Looks good, I assume this is to sort out this one

[Fail] [sig-network] Services [It] should be able to change the type from ExternalName to ClusterIP [Conformance]
/workspace/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/network/service.go:1333

jayunit100:
filed #194 as a follow on

PR #213, opened 2022-01-14
fail kpng local up if compilation is broken

### LINKED ISSUE #212: kpng local up silent fail

jayunit100:
essential MR to fix the fact that kpng-local-up fails silently ... 

fixes #212


dougsland (approved):
/lgtm

PR #218, opened 2022-01-16
go1.18 improvements & diffstore2

mcluseau:
Do not merge before go 1.18

This contains the diffstore2 (did not found a better name) with Example functions as unit tests.

I'll follow up refactoring `backends/nft` to use it as an operational example.

@hanamantagoudvk @jayunit100 @knabben


dougsland:
Nice one! Just a nit comment: https://github.com/kubernetes-sigs/kpng/blob/master/go-mod-tidy-all#L19 would require to be bump as well if 1.18 is the latest minimum version for kpng.

client/diffstore2/buffer-leaf.go
```diff
@@ -0,0 +1,34 @@
+package diffstore2
+
+import (
+	"bytes"
+    "constraints"
```
knabben, line 5: gofmt

mcluseau: yeah it's currently broken in my setup

dougsland:
@mcluseau how about ship this "module" as separate project? It will reduce the complexity of the kpng code tree and users will always have a option to glue whatever btree implemention they want.

mcluseau:
@dougsland it's already just a tool people can use if they want but for now, we don't know how closely integrated this will be with kpng backends. We should let this solidify before thinking about splitting. (I supposed you were talking about the `diffstore2` package)

dougsland:
> @dougsland it's already just a tool people can use if they want but for now, we don't know how closely integrated this will be
> with kpng backends. We should let this solidify before thinking about splitting. (I supposed you were talking about the
> `diffstore2` package)

@mcluseau Agreed, make sense to keep it for now. In the future we can bring the topic again if necessary.

Issue #178, opened 2022-01-04
add hanamant, vivek, lars to root owners


jayunit100:
since hanamant vivek and lars have all been involved for a long time and worked on some important core features and are domain experts for us, lets add them to root owners file

PR #231, opened 2022-01-21
adding myself, vivek root owners

hanamantagoudvk:
#178


hanamantagoudvk:
Since lars is not yet members of k8s-sigs , i have not added his name yet to owner file. Once he is a members, we will add him here.

PR #239, opened 2022-01-29
Multiport fix when targetport is a string

### LINKED ISSUE #187: [e2e debug] should serve multiport endpoints from pods
What the test does -
Create a service with two ports but no Pods are added to the service yet.  The service MUST run and show empty set of endpoints. Add a Pod to the first port, service MUST list one endpoint for the Pod on that port. Add another Pod to the second port, service MUST list both the endpoints. Delete the first Pod and the service MUST list only the endpoint to the second Pod. Delete the second Pod and the service must now have empty set of endpoints.

anusha94:
- update services.proto to include endpoint map
- update slice event handler to populate this map
- modify dnat rules to fetch port from the map when svc target port is a
  string

Signed-off-by: Neha Lohia <nehapithadiya444@gmail.com>
Signed-off-by: Vivek Thrivikraman <vivek.thrivikraman@est.tech>

Fixes #187


jayunit100:
imo merge + iterate -> can lift into the brain end later

jayunit100:
imo merge + iterate -> can lift into the brain end later

jayunit100:
ok to merge this and itereate forward @VivekThrivikraman-est ? lets make sure to add all commiter emails to it for folks that contributed to this

jayunit100:
can iterate on a fancier solution later.
this just manipulates the data locally which is fine for getting multiport e2es to pass  
/lgtm
/approve

PR #242, opened 2022-01-30
EndpointPortMap update fix

### LINKED ISSUE #241: comment the slice-event-handler OnAdd logic
this code in slice-event-handler. OnAdd... 
```
		for _, port := range eps.Ports {
			ep := map[string]int32{
				*port.Name: *port.Port,
			}
			info.Endpoint.EndpointPortMap = ep
		}
```
is confusing , i.e. what is this final port that were prioritizing :) , it recently was written to fix a multiport sessinoAffinity e2e failure, i believe... but, lets just make sure we comment the specific logic of this .... 

/assign @anusha94 @nehaLohia27

anusha94:
Fixes #241 

Instead of choosing only the last port, we need to assign all `port names: numbers` to the endpoint port map.

### DIFF
--- a/server/jobs/kube2store/slice-event-handler.go
+++ b/server/jobs/kube2store/slice-event-handler.go
@@ -72,12 +72,11 @@ func (h sliceEventHandler) OnAdd(obj interface{}) {
 			info.Endpoint.AddAddress(addr)
 		}
 
+		epMap := make(map[string]int32, len(eps.Ports))
 		for _, port := range eps.Ports {
-			ep := map[string]int32{
-				*port.Name: *port.Port,
-			}
-			info.Endpoint.EndpointPortMap = ep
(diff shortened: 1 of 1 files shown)


server/jobs/kube2store/slice-event-handler.go
```diff
@@ -72,12 +72,11 @@ func (h sliceEventHandler) OnAdd(obj interface{}) {
 			info.Endpoint.AddAddress(addr)
 		}
 
+		epMap := make(map[string]int32, len(eps.Ports))
 		for _, port := range eps.Ports {
-			ep := map[string]int32{
-				*port.Name: *port.Port,
-			}
-			info.Endpoint.EndpointPortMap = ep
+			epMap[*port.Name] = *port.Port
```
jayunit100, line 77: hahhaa ya this makes sense:)

PR #245, opened 2022-02-01
e2e_test: Made kind ipv6 core dns patch default when using IPv6

qabpea:
Included some upstream updates from e2e tests
Added small description regarding "not-kpng"


dougsland (approved):
/lgtm
/approve

PR #249, opened 2022-02-05
conntrack module

mcluseau:


client/plugins/conntrack/conntrack_linux.go
```diff
@@ -0,0 +1,68 @@
...
+	return parameters
+}
+
+func protoStr(proto v1.Protocol) string {
+	return strings.ToLower(proto.String())
+}
+
+// IsClearConntrackNeeded returns true if protocol requires conntrack cleanup for the stale connections
+func IsClearConntrackNeeded(proto v1.Protocol) bool {
+	return proto == v1.Protocol_UDP || proto == v1.Protocol_SCTP
```
knabben, line 67: why do need to cleanup SCTP conntrack?

client/plugins/conntrack/conntrack_linux.go
```diff
@@ -0,0 +1,68 @@
...
+	"k8s.io/utils/exec"
+	utilnet "k8s.io/utils/net"
+
+	v1 "sigs.k8s.io/kpng/api/localnetv1"
+)
+
+var execer = exec.New()
+
+func setupConntrack() {
+	// TODO
```
knabben, line 17: do we need to setup anything before calling the conntrack module, or this is only a placeholder?

client/plugins/conntrack/conntrack_linux.go
mcluseau: this code is from upstream ;)

client/plugins/conntrack/conntrack_linux.go
mcluseau: there are things to do yes (upstream: cmd/kube-proxy/app/conntrack.go)

backends/nft/register.go
```diff
@@ -25,7 +27,12 @@ func (b *backend) Sink() localsink.Sink {
 	sink := fullstate.New(&b.cfg)
 
 	PreRun()
-	sink.Callback = Callback
+
+	ct := conntrack.New()
+	sink.Callback = fullstatepipe.New(fullstatepipe.ParallelSendSequenceClose,
+		Callback,
+		ct.Callback,
```
jayunit100, line 34: this is a new "mini backend" that is executed after the original , NFT backend

client/plugins/conntrack/conntrack.go
```diff
@@ -0,0 +1,60 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	ct.flows.Done()
+
+	for _, item := range ct.flows.Deleted() {
+		cleanupFlowEntries(item.Value().Get())
```
VivekThrivikraman-est, line 58: I think cleaning flow entries should also be done when endpoint changes from no endpoint to >=1  endpoints as there could be stale entries then.
We should also be checking the ready state of endpoints , else there are chances of stale connections(even if we take care of transition from 0 to 1 endpoints)

backends/nft/register.go
```diff
@@ -25,7 +27,12 @@ func (b *backend) Sink() localsink.Sink {
 	sink := fullstate.New(&b.cfg)
 
 	PreRun()
-	sink.Callback = Callback
+
+	ct := conntrack.New()
+	sink.Callback = fullstatepipe.New(fullstatepipe.ParallelSendSequenceClose,
```
VivekThrivikraman-est, line 32: I think we should be calling the tasks in serial order 1)Apply the rules 2)Delete the stale endpoints.

client/plugins/conntrack/conntrack_linux.go
VivekThrivikraman-est, line 67: I think SCTP can be actually removed

PR #250, opened 2022-02-07
test_e2e: fix to make ipv6 tests run correctly (or detect incompatible sysctl settings)

qabpea:
Added code that verify the host network settings and also code that set them if ci/cd is set.

Made delete of clusters used for test default behaviour
Small changes in ginko settings in test_run
Prepared for changes in how kind cluster is created


qabpea:
Looks like ci/cd expects the cluster to not be removed. I will push an update that does not remove the cluster when ci-cd flag is set

dougsland:
late for the game but lgtm

Issue #247, opened 2022-02-05
Serious QA issues: we've merged code that doesn't even compile


mcluseau:
Missing comma:
https://github.com/kubernetes-sigs/kpng/blob/master/client/serviceevents/service-events_test.go#L52

Also, unit tests are not passing.

I think it is the minimum we can require before merging something. We should have those checks in github's pre-merge conditions.

/good-first-issue

yankay:
@mcluseau I try to fix the environment. But there are still some unit test cannot pass

yankay:
HI , @mcluseau . There is the https://github.com/kubernetes-sigs/kpng/pull/251 to fix the issue . Would you please review it .

PR #251, opened 2022-02-08
unit test should exit 1 if fail

yankay:
When unit test run fail，the PR cannot be merge.

to fix https://github.com/kubernetes-sigs/kpng/issues/247


yankay:
Github action do not support Golang 1.18. So there are some package cannot be compiled

mcluseau:
@yankay thanks for this initial work :)

@nehaLohia27 this touches github actions

@knabben this touches userspace

@hanamantagoudvk this touches serviceevents tests and it's weird that we find the "want:" line in the output comment of the example test.

knabben (commented):
/lgtm

Issue #261, opened 2022-02-18
pipeSink


jayunit100:
@VivekThrivikraman-est going build a pipe sink !  Heres how it will work ...

```
	return filterreset.New(
		pipeSink.New(
			decoder.New(s), // pipe input 1
			conntrack.New(s), // pipe input 2
					// ...
		)
	)
```

Then we can use it to push events into conntrack cleanup module.  

@anusha94  @VivekThrivikraman-est 

- will be easier then the fullstate pipe bc only one strategy  flowing events all the way through.
- fullstate had parallel/serial/parallelSend-serialClose

PR #266, opened 2022-02-25
Pipe for filterreset and stale conntrack handling for iptable

VivekThrivikraman-est:
Currently stale conntrack entries for iptables are not handled and hence conntrack related test cases are failing.

Fixes: #261


jayunit100:
/lgtm
/approve

This is required so that we can pass the 1 final failing IP TAbles testttttttttttt

Issue #282, opened 2022-05-25
Add make target for grpc compilation


jayunit100:
we need to:
- document how to build our GRPC stubs which live in https://github.com/kubernetes-sigs/kpng/blob/master/api/localnetv1/services.proto 
```
apt install protoc

protoc -I=./protobuf 
--go_out=plugins=grpc:./protobuf 
./protobuf/kpng.proto

... 
```

- make a hack/generate-grpc.sh script
- make a hack/verify-grpc.sh script
- add a ci job that runs "generate-grpc.sh" and then compares (diffs) content w/ verify-rpc.sh -> fails if different

/good-first-issue

jayunit100:
we need to:
- document how to build our GRPC stubs which live in https://github.com/kubernetes-sigs/kpng/blob/master/api/localnetv1/services.proto 
```
apt install protoc

protoc -I=./protobuf --go_out=plugins=grpc:./protobuf ./protobuf/kpng.proto

... 
```

- make a hack/generate-grpc.sh script
- make a hack/verify-grpc.sh script
- add a ci job that runs "generate-grpc.sh" and then compares (diffs) content w/ verify-rpc.sh -> fails if different

/good-first-issue

kanha-gupta:
Hi i would like to contribute
i am a newbie

yashsingh74:
Open a PR- https://github.com/kubernetes-sigs/kpng/pull/285

PR #285, opened 2022-06-06
Add a doc and script for grpc compilation

yashsingh74:
This PR has the document about how to build the gRPC stubs and the script to generate them.

Fixes for #282


mcluseau:
hi, I think the protobuf subdir will break the package we want. To me the command should be something like

    protoc -I ./ --go_out=paths=source_relative:. --go-grpc_out=paths=source_relative:. $(find -name '*.proto')

yashsingh74:
> hi, I think the protobuf subdir will break the package we want. To me the command should be something like
> 
> ```
> protoc -I ./ --go_out=paths=source_relative:. --go-grpc_out=paths=source_relative:. $(find -name '*.proto')
> ```

@micahhausler 
I ran that command from the home directory of this project and gives me the following result-
```
protoc -I ./ --go_out=paths=source_relative:. --go-grpc_out=paths=source_relative:. $(find -name '*.proto')
find: illegal option -- n
usage: find [-H | -L | -P] [-EXdsx] [-f path] path ... [expression]
       find [-H | -L | -P] [-EXdsx] -f path [path ...] [expression]
Missing input file.
```

rajaskakodkar:
```
protoc -I ./ --go_out=paths=source_relative:. --go-grpc_out=paths=source_relative:. $(find . -name '*.proto')
```
cc: @yashsingh74 this worked for me. Also, try to fix the make target for modd here https://github.com/kubernetes-sigs/kpng/blob/master/modd.conf.template#L5

@jayunit100  @knabben  do we need scripts under hack/ or can we make the make target with modd work?

anusha94:
> @jayunit100 @knabben do we need scripts under hack/ or can we make the make target with modd work?

And maybe need another Good First Issue to have dependent targets that will install `protoc`?
Until then, having the scripts makes sense. But maybe not once we get `make modd` working

aahemm:
Hi @yashsingh74,
I am kind of a beginner in this topic. I have gone through 
your commits and have a couple of questions: a) Shouldn't the verify script exit with a non zero exit code when the pb files differ? It will make the pipeline fail and I think this is the desired behavior. b) Where do we generate the files in the `examples` directory? I mean the files that are `diff`ed against actual pb files.

examples/protobuf/services.pb.go
```diff
@@ -0,0 +1,1992 @@
+// Copyright 2021 The Kubernetes Authors.
```
anusha94, line 1: we should not check-in `examples` dir. This is only to verify the diff with the generated one vs the checked-in one.
Also, I think a better name to this dir would be something like `verify` or `validate`

yashsingh74: Where should I keep the verify dir ?

jayunit100: hack/verify... is fine @yashsingh74

jayunit100:
thank you for this yash ! lets try to merge it this wk

jayunit100:
i think im ok w/ this ? it doesnt break anything and its a step in the right direction.  that said i think probably we should work together on this so that its clear what the trajectory is.

if nobody objects ill merge it tomorrow.

jayunit100:
@yashsingh74 can you rebase to one commit please?

mcluseau:
I'm approving the tests to run, but am I the only one who sees the github button to do that? @jayunit100

mcluseau:
@yashsingh74 thanks for this PR, it even has doc so it's better than what I do :p a little scripts needs its `chmod +x` though: test fails with `./hack/verify/generate-grpc.sh: Permission denied`

yashsingh74:
@mcluseau I have make the script executable. Please try around once more.

mcluseau:
@yashsingh74 ok now it runs, but still fails (protoc dependency missing) https://github.com/kubernetes-sigs/kpng/runs/7427413370?check_suite_focus=true

jayunit100:
unit tests + grpc yay ! ok , lets get them passing now @yashsingh74

hack/verify/verify-grpc.sh
```diff
@@ -0,0 +1,29 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Verifying the protobuf
+
+function verify {
+    diff -q api/localnetv1/verify-protobuf/services.pb.go api/localnetv1/services.pb.go && diff -q api/localnetv1/verify-protobuf/services_grpc.pb.go api/localnetv1/services_grpc.pb.go || { echo "File doesn't match"; exit 1; }
```
jayunit100, line 20: can we print the entire diff?

.github/workflows/grpc_tests.yml
```diff
@@ -0,0 +1,35 @@
...
+        go version
+        go install golang.org/x/lint/golint@latest
+        sudo apt-get install -y libnl-3-dev libnl-genl-3-dev
+
+    - name: checkout
+      uses: actions/checkout@v2
+     
+    - name: run grpc verify tests
+      run: |
+        ./hack/verify/generate-grpc.sh
```
jayunit100, line 33: `./hack/verify/generate-grpc.sh && ./hack/verify/verify-grpc.sh` ?

jayunit100:
talked w/ @mcluseau @yashsingh74 

Lets 
- add Dockerfile into hack/ 
- use *that* to generate the protobuf
- use `ubuntu:latest` probably as the Dockerfile? 
- file a follow on issue to make as many jobs as we can, 
```
jobs:
  test:
    name: run grpc verify tests
```
run in a container (the same one ideally?) so that theres 100% parity between the ci jobs build env and developer env.

jayunit100:
@anusha94 do you want to help yash finish this one ?

PR #291, opened 2022-06-27
added comment for serverProxyMode

### LINKED ISSUE #289: Add a quick comment/link back to the service.go e2e
The service.go e2e queries the kube proxy for the "type" of proxy as part of the way it tests... 

https://github.com/kubernetes/kubernetes/blob/b1e2af45e4ff126fcbcffa2e66ed36502103be7b/test/e2e/network/service.go#L3911

In cases where you dont have a kube proxy running, these tests dont run at all.. i.e. they are backend specific

We should comment why we run "servceProxyMode" for the future so that people looking at our backends understand its just for backwards compatibility w/ upstream e2es

Lets add a comment to ... 

https://github.com/kubernetes-sigs/kpng/blob/master/backends/ipvs-as-sink/ipvs.go#L287

chetak123:
Fixes #289 

I have added comment for serverProxyMode in ipvs.go file.

### DIFF
--- a/backends/ipvs-as-sink/ipvs.go
+++ b/backends/ipvs-as-sink/ipvs.go
@@ -284,6 +284,9 @@ func (s *Backend) SetUpHttpListen() error {
 	return <-errCh
 }
 
+
+// ServeProxyMode used for creating the type of proxy which can be tested by proxyMode in service.go e2e test 
+// i.e for backward compatbility 
 func (s *Backend) ServeProxyMode(errCh chan error) {
 	//TODO Get Bind address config. Time being leave it empty, kernel will choose loopback adress 127.0.0.1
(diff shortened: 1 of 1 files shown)


backends/ipvs-as-sink/ipvs.go
```diff
@@ -284,6 +284,9 @@ func (s *Backend) SetUpHttpListen() error {
 	return <-errCh
 }
 
+
+// ServeProxyMode used for creating the type of proxy which can be tested by proxyMode in service.go e2e test
```
jayunit100, line 288: what do we mean by "type of proxy" here? I've already forgotten why we had to port this in from the original codebase in kubernetes/kubernetes

mcluseau:
Unresolved conversation since jul 24, closing until someone reopens.

PR #306, opened 2022-07-24
Include example in go.work

uablrek:
The `examples/` items should be included in `go.work`, else they can't be built.

Also use $VERSION in binaries. I not set the version defaults to the output from;
```
git describe --dirty --tags
```

I myself becomed very annoyed if I can't get the version of programs I am using and I assume I am not alone.

### DIFF
--- a/hack/go-build
+++ b/hack/go-build
@@ -1,2 +1,3 @@
 set -ex
-CGO_ENABLED=1 go build -trimpath -o dist $(hack/go-list-local-mods)
+test -n "$VERSION" || VERSION=$(git describe --dirty --tags)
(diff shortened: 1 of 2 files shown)


hack/go-build
```diff
@@ -1,2 +1,3 @@
 set -ex
-CGO_ENABLED=1 go build -trimpath -o dist $(hack/go-list-local-mods)
+test -n "$VERSION" || VERSION=$(git describe --dirty --tags)
```
mcluseau, line 2: I'd say `: ${VERSION:=$(git describe --dirty --tags)}` :)

PR #307, opened 2022-07-24
discovery.k8s.io/v1beta1 -> v1

uablrek:
The v1beta1 version is removed in K8s v1.25+
```
W0724 12:25:51.279069     951 warnings.go:70] discovery.k8s.io/v1beta1 EndpointSlice is deprecated in v1.21+, unavailable in v1.25+; use discovery.k8s.io/v1 EndpointSlice
```

The `Topology` field is removed. I don't know the impact, I just removed it.

### DIFF
--- a/server/jobs/kube2store/slice-event-handler.go
+++ b/server/jobs/kube2store/slice-event-handler.go
@@ -17,7 +17,7 @@ limitations under the License.
 package kube2store
 
 import (
-	discovery "k8s.io/api/discovery/v1beta1"
+	discovery "k8s.io/api/discovery/v1"
 	"k8s.io/klog/v2"
 
 	localnetv1 "sigs.k8s.io/kpng/api/localnetv1"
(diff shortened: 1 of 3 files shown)


server/jobs/kube2store/slice-event-handler.go
```diff
@@ -51,15 +51,10 @@ func (h sliceEventHandler) OnAdd(obj interface{}) {
 			Namespace:   eps.Namespace,
 			ServiceName: serviceName,
 			SourceName:  eps.Name,
-			Topology:    sliceEndpoint.Topology,
```
astoycos, line 54: Can we remove that field from the grpc API @mcluseau? 

https://github.com/kubernetes-sigs/kpng/blob/master/api/localnetv1/services.proto#L168

astoycos:
/lgtm barring my one question ^ 

Thanks!

server/jobs/kube2store/slice-event-handler.go
mcluseau: we have to assess the use of topology keys, ie for local endpoint detection, I don't have that in brain-cache

uablrek: But that field is removed, I don't have that in brain-cache either

mcluseau: yes but we can't just delete the line in reaction :) ie, topology is used to detect local endpoints. I'll check that.

uablrek:
Since I wrote https://github.com/kubernetes-sigs/kpng/issues/317 and will close this PR since it's not as trivial as I thought.

mcluseau:
@uablrek a few changes in core before and this will be mergable, @astoycos is right about the need to do something with the topology info, but we'll still need it to address TopologyAwareHints.

Issue #290, opened 2022-06-24
mutating service types doesnt trigger ServiceEvents 


jayunit100:
@hvkk8s is looking at our failing IPVS tests 

```
should be able to update service type to NodePort listening on same port number but
different protocols
```

The bug is:
- make a ClusterIP
- update it to a NodePort

And the 2nd thing doesnt get triggered in our events, so kpng backend never actually updates the IPVS service to open up that nodeport.

PR #287, opened 2022-06-07
Endpoint scopes

mcluseau:
Introduce the notion of endpoint scopes to allow generic filtering with a simple node-local model.

Includes 2 first scopes: Internal and External.

Depends on #286 
Should fix #277

PR #314, opened 2022-07-27
Change endpoint key and use anonymous set depending on presence of PodName

gauravkghildiyal:
Replacing a service could (and can) result in the following sequence of actions:

1. ipvssink.Backend.SetEndpoint(IP = 10.0.0.1, Key = A)
2. ipvssink.Backend.SetEndpoint(IP = 10.0.0.1, Key = B)
3. ipvssink.Backend.DeleteEndpoint(IP = 10.0.0.1, Key = A)

Ideally, once these operations are complete, one would expect that the IP 10.0.0.1 should exist after the completion of these actions, but since duplicates are ignored while adding Real Servers to an IPVS Virtual Server, this results in the IP being completely removed. 

### Solution

Refer the discussion below on how we arrived at this solution.

#### Discarded Solution
The initial proposal was to keep a count of the usages that an IP has in a service and only remove it from the IPVS Virtual Server when its usage drops to zero.

#### Final solution
We have now changed the `key` for an endpoint such that it changes less often (when the configuration of the endpoint changes). The effect this has is that if the key remains the same, then instead of deleting the old endpoint and creating a new one, we simply update the existing endpoint - meaning no delete event is triggered.

The key calculation relies on the name of the pod corresponding to the endpoint. There is a case when the pod name might be absent and to deal with this scenario, we introduce another diffing set. Changes to this new set result in sending the delete events prior to the update events.

This also closes https://github.com/kubernetes-sigs/kpng/issues/290 by comparing ports properly to propagate new IPPort additions


gauravkghildiyal:
Besides the already existing flakes, it looks good to be reviewed from my side. 

`should be able to update service type to NodePort listening on same port number but
different protocols` is no longer failing even after multiple runs.

Other flakes can be investigated separately.

mcluseau:
I'm worried about the changes in service-events. It's putting a lot of business logic in the client, which is not what we want, and endpoints deletions sent by the local API should always be legitimate deletions. Also, since we always get endpoint deletions before the service deletions, there will always be a "valid usage" at the time of endpoint deletion. Are those changes in service-events really fixing something? If yes, it's probably the wrong place to fix it, as the kpng local API shouldn't send illegitimate endpoint deletions.

gauravkghildiyal:
I understand your valid concern here. 

Before I get into proposing what could be a possible solution, let me explain the scenario which results in an inconsistency, so that we're clear on the problem that I'm trying to address (There's a more lengthy version of the details at the end).

### The problem
As per the order of events defined here https://github.com/kubernetes-sigs/kpng/blob/388d3c3c04781ee48bd5e39ff72b001246461937/server/jobs/store2localdiff/store2localdiff.go#L118-L121, backends could result in receiving events in the following order:
1. `SetEndpoint(IP = 10.0.0.1, Ports = [TCP:8080], SomeUniqueIdentifier = AAAA)`
2. `SetEndpoint(IP = 10.0.0.1, Ports = [TCP:8080, UDP:8081], SomeUniqueIdentifier = BBBB)`
3. `DeleteEndpoint(IP = 10.0.0.1, Ports = [TCP:8080], SomeUniqueIdentifier = AAAA)`

> kpng local API shouldn't send illegitimate endpoint deletions.

So overall, kpng local API doesn't send illegitimate deletions, it's just the relative ordering of additions and deletions which causes the problem.

### Possible solutions
1. The first solution is the currently implemented one, where the backend takes care of maintaining the frequency of the endpoint, and only delete it when the frequency drops to zero.
2. The second solution is about directly tackling the source of confusion - which I hadn't explored earlier and hence ended up with solution 1 first. Basically, _do deletes before updates_:

```
	count += w.SendDeletes(localnetv1.Set_EndpointsSet)
	count += w.SendDeletes(localnetv1.Set_ServicesSet)
	count += w.SendUpdates(localnetv1.Set_ServicesSet)
	count += w.SendUpdates(localnetv1.Set_EndpointsSet)
```

### Request for input
The 2nd solution is obviously smaller and cleaner, but I'd like to confirm if there's any reason why we do the updates first and then the deletes. If there's no compelling reason, I'll just change the PR to solution 2. Otherwise, we definitely need to check for this on the backend - which is the current solution.

Please let me know your thoughts here @mcluseau.

---

### Steps to reproduce and more detailed explanation

1. Create any sample deployment with **atleast one replica**: 
`kubectl apply -f deployment.yaml`

    <details>
      <summary>deployment.yaml</summary>
      <pre>
    apiVersion: apps/v1
    kind: Deployment
    metadata:
      labels:
        app: echoserver
      name: echoserver
    spec:
      replicas: 1
      selector:
        matchLabels:
          app: echoserver
      template: 
        metadata:
          labels:
            app: echoserver
        spec: 
          containers:
          - name: echo
            image: gcr.io/kubernetes-e2e-test-images/echoserver:2.2
            ports:
            - name: echoserver-tcp
              containerPort: 8080
              protocol: TCP
            - name: echoserver-udp
              containerPort: 8081
              protocol: UDP
      </pre>
    </details>

    #### Result
    Lets assume the ClusterIP assigned to the pod is `10.0.0.1`

3. Create a service that just listens on **one port**:
`kubectl apply -f service.yaml`

    <details>
      <summary>service.yaml</summary>
      <pre>
    apiVersion: v1
    kind: Service
    metadata:
      name: echoserver-svc
    spec:
      type: ClusterIP
      selector:
        app: echoserver 
      ports:
      - name: tcp
        port: 80 
        protocol: TCP
        targetPort: echoserver-tcp 
      </pre>
    </details>

    #### Result
    After this step, we'll firstly receive the event for the service creation. After that we'll receive an event for the endpoint addition - something like this:
    `SetEndpoint(IP = 10.0.0.1, Ports = [TCP:8080], SomeUniqueIdentifier = AAAA)`
    All of this is "valid-usage" and everything's good so far.

5. Update the same service to listen on **two ports**: 
`kubectl apply -f modified-service.yaml`

    <details>
      <summary>modified-service.yaml</summary>
      <pre>
    apiVersion: v1
    kind: Service
    metadata:
      name: echoserver-svc
    spec:
      type: ClusterIP
      selector:
        app: echoserver 
      ports:
      - name: tcp
        port: 80 
        protocol: TCP
        targetPort: echoserver-tcp 
      - name: udp
        port: 81
        protocol: UDP
        targetPort: echoserver-udp
      </pre>
    </details>
    
    #### Result
    Things become problematic after this step. Since this step just updates the endpoint, the ClusterIP for the pod remains the same - just another port has been added to it. This update will result in the following sequence of events being generated.
    * `SetEndpoint(IP = 10.0.0.1, Ports = [TCP:8080, UDP:8081], SomeUniqueIdentifier = BBBB)`
    * `DeleteEndpoint(IP = 10.0.0.1, Ports = [TCP:8080], SomeUniqueIdentifier = AAAA)`
   Now this is still valid usage in the sense that the service wasn't deleted - just the endpoints were updated and that too correctly. But since the final actual proxiers like `ipvs`, `nft` don't allow duplicates to be added, the `DeleteEndpoint` event above results in the overall deletion of that IP:Port pairs for TCP.

mcluseau:
@gauravkghildiyal thanks for the detailed analysis.

There's at 3rd option, and it is an issue I suspected since endpoint go more complex: currently, the key for an endpoint is its full hash, which means it's added+deleted when it changes. So, the issue is, as endpoints get more mutable info, that this simple model is not precise enough and should be changed. I think it would make sense to change the endpoint's key to `namespace+name+podName`.

This should fix the issue you're describing as there wouldn't be a delete anymore. Backends not relying on the fullstate helper would have to handle updates, but it's probably already nearly the case.

Does it makes sense? :)

gauravkghildiyal:
@mcluseau Thanks for the explanation. :) I think I kinda understood what you meant but would still like to confirm.

So previously, the `hash` and `key` for the endpoint were the same (to be precise, the key was derived from the hash, but only depended on the hash so can be considered equal for the time being). The logic used to control the deletion depended on the `hash`

Now, I've changed the `hash` such that it based on the `namespace` + `name` + `podName`. The key part is still the same as before - it will change if the endpoint struct changes. Thus, when an update happens, there will no longer be any delete events triggered, but this would also mean that the btree still has the old endpoint.

I've made changes with the assumptions stated above. Please let me know if incase I misunderstood something. 

Thanks for the patience!

mcluseau:
@gauravkghildiyal nope, hash must stay the same (hashing the full object) because it is use to detect change in the endpoints data, we only want to change the way the key is computed in `store2localdiff.go` to something like (I think Hostname is reliable enough now):

```golang
			key := append(make([]byte, 0, len(key)+1+len(ei.Endpoint.Hostname)), key...)
			key = append(key, '/')
			key = append(key, []byte(ei.Endpoint.Hostname)...)
```

gauravkghildiyal:
@mcluseau Ah okay things make perfect sense now! I was wrongly assuming you meant the hash and key inside the `proxystore.go` file.

I just have some (probably final :) )doubts about the choice of the new key:

1. Keeping the key as `serviceKey + hostname` seems problematic since `hostname` is only set when it's explicitly specified by the user. Thus, hostnames for two pods of the same service would result in conflicts for the key and one would overwrite the other. From what I could understand, we do want to have uniqueness between two endpoint keys right?

2. I thought about keeping the key as `serviceKey + podName` as per your earlier suggestion, but that too might be problematic for the case when the `podName` is not available in the Endpoints (or EndpointSlice) objects - which would again lead to the conflicts between keys. One example I could think of is the case with [Services without selectors](https://kubernetes.io/docs/concepts/services-networking/service/#services-without-selectors) where the user manually adds IPs to endpoints.

So with these thoughts, I was thinking we could keep the key as `serviceKey + hash(IPs)`. This to be precise:
```golang
ipHash := serde.Hash(ei.GetEndpoint().GetIPs())
key := append(make([]byte, 0, len(key)+1+64/8*2), key...)
key = append(key, '/')
key = strconv.AppendUint(key, ipHash, 16)
```

Does this seem reasonable? Sorry if I still lack some context somewhere.

mcluseau:
@gauravkghildiyal k8s is a big context and we probably all lack some of it :p I'm happy to help people get in this project, the more we are to understand it at this level, the better :)

The hash of IPs was an idea but the problem is that IPs can be shared between multiple pods if they are on host network, and with all the complexity introduced by named ports, it can be a valid use case (ie: 2 host-network pods with a different "metrics" port).

I based my evaluation about hostname on the definition in EndpointSlice.Endpoint (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#endpoint-v1-discovery-k8s-io) which states *this field may be used by consumers of endpoints to distinguish endpoints from each other*. The pod name should also be available in this object's `targetRef` too, and I'm ok with that, but using it means you'll have to add a field (`string TargetRef`?) in `localnetv1.Endpoint`.

But I didn't think about the services without selectors case (so you see, no one has the full context in mind :sweat_smile: ). What about using the hash of `addresses+ports` only when the hostname is empty?

gauravkghildiyal:
> k8s is a big context and we probably all lack some of it

True :)

---

> I'm happy to help people get in this project,

Yes you have been very helpful and patient. Thanks for that

---

> The hash of IPs was an idea but the problem is that IPs can be shared between multiple pods if they are on host network,...

Oops didn't think of this. Thanks for pointing out.

---

> I based my evaluation about hostname on the definition in EndpointSlice.Endpoint...

Oh the comment seems a bit misleading to me.
I think hostname is very unreliable since in most cases, it is almost never set. (Condition for setting it being that the user explicitly specifying the `hostname` and `subdomain`( = service name) in the pod spec) :

Ref:

[kubernetes/kubernetes/pkg/controller/endpointslice/utils.go](https://github.com/kubernetes/kubernetes/blob/d046a58de435a07cf5d0d2d0879c3b21dbb52cca/pkg/controller/endpointslice/utils.go#L76-L78)
```golang
	if endpointutil.ShouldSetHostname(pod, service) {
		ep.Hostname = &pod.Spec.Hostname
	}
```

[kubernetes/kubernetes/blob//pkg/controller/util/endpoint/controller_utils.go](https://github.com/kubernetes/kubernetes/blob/d046a58de435a07cf5d0d2d0879c3b21dbb52cca/pkg/controller/util/endpoint/controller_utils.go#L162-L164)
```golang
// ShouldSetHostname returns true if the Hostname attribute should be set on an
// Endpoints Address or EndpointSlice Endpoint.
func ShouldSetHostname(pod *v1.Pod, svc *v1.Service) bool {
	return len(pod.Spec.Hostname) > 0 && pod.Spec.Subdomain == svc.Name && svc.Namespace == pod.Namespace
}
```

---

> What about using the hash of addresses+ports only when the hostname is empty?

Wouldn't using the hash of address+ports basically be equivalent to what exists now and circle back to the problem in https://github.com/kubernetes-sigs/kpng/pull/314#issuecomment-1200304152 about adding new service ports?

Finding a new key that satisfies our needs does not seem to be trivial afterall.
Does `ServiceKey + IPs + PodName`  work?:
* Other than the case of service without selectors, `PodName` should guarantee uniqueness.
* For the case of service without selectors where the PodName is empty (and hence the same for all such cases), the `IPs` should guarantee uniqueness since in this case I don't think we have the concept of "named ports" (as we have no selectors)

Again, I'm not sure of the validity but if all this is just over complicating things, then what about changing the order of sending diffs? i.e. Deletes before Updates? Since the current hash/key strategy seems to have worked so far, we can continue using that with this change. 

Let me know your thoughts whenever you have some time. Thanks!

mcluseau:
let's first note that we're ok with the pod name case, to be stored in `localnetv1.EndpointInfo`.

In the service without selector case, I can totally do that:
```yaml
---
apiVersion: v1
kind: Service
metadata:
  name: metrics
spec:
  type: ClusterIP
  ports:
  - name: metrics
    port: 1234
---
apiVersion: v1
kind: Endpoints
metadata:
  name: metrics
subsets:
- addresses:
  - ip: 10.0.0.1
  ports:
  - name: metrics
    port: 1001
- addresses:
  - ip: 10.0.0.1
  ports:
  - name: metrics
    port: 1002
```

which is translated to:
```
$ kubectl -n default get ep metrics
NAME      ENDPOINTS                     AGE
metrics   10.0.0.1:1001,10.0.0.1:1002   6s
$ kubectl -n default get endpointslices -l kubernetes.io/service-name=metrics
NAME            ADDRESSTYPE   PORTS   ENDPOINTS   AGE
metrics-p4q8w   IPv4          1001    10.0.0.1    4m15s
metrics-q99v2   IPv4          1002    10.0.0.1    4m15s
```

The problem with delete-before-update is that a can service go through a "0 endpoints" state, while it may still have endpoints. It's easier to manage the client side if we have first have the information that endpoints exist for the service (note: clients using fullstate are not affected). We may be forced to accept this degradation for services without selector though.

The current hash-as-key strategy worked so far, but it's showing its limits ie with the introduction of endpoint score (#287 ), so the need for a better key is here anyway.

tldr, I'm mentally drafting this could work:
- introduce a 2nd endpoint set in watchstate (I'll propose a PR for this because I need to write code to wrap my head around this :o )
- the key always has the prefix `namespace/service-name/`
- if there's a pod name: add it to the key
- else, use the current hash-as-key strategy, but store the result in the second endpoint set
- send updates/deletes in this order:
- ```golang
count += w.SendUpdates(localnetv1.Set_ServicesSet, 0)
count += w.SendDeletes(localnetv1.Set_EndpointsSet, 1)
count += w.SendUpdates(localnetv1.Set_EndpointsSet, 0)
count += w.SendDeletes(localnetv1.Set_EndpointsSet, 0)
count += w.SendUpdates(localnetv1.Set_EndpointsSet, 1)
count += w.SendDeletes(localnetv1.Set_ServicesSet, 0)
```

gauravkghildiyal:
Thanks @mcluseau . I understand now why IPs as hashes will still be problematic and the need for update-before-delete (to have lesser downtimes when services change by preventing the 0 endpoints state when possible).

Your suggestion also makes sense i.e. degrading to the delete-before-update case for the endpoints which don't have a pod name.

I'm guessing you do plan on merging https://github.com/mcluseau/kpng/commit/0503286c2ed7da82fdf141fb5150a3e5b8ca16c3 eventually right? In which case I'll not include that in this PR. Thanks!

mcluseau:
you can include the commit in this PR, so the discussion is linked to it, and you can also check that it's actually solving the issue before merging :)

gauravkghildiyal:
I see. Will do the needful in some time, verify, and get back later.

Thanks a lot!

gauravkghildiyal:
Ready for review. Looks good after manual tests.

mcluseau:
great then :)

/lgtm
/approve

PR #318, opened 2022-07-30
migrate to discovery/v1

### LINKED ISSUE #262: Add TopologyAwareHints to the KPNG brain
We use the old topology API currently: https://github.dev/kubernetes-sigs/kpng/blob/990877f589ef015a24562a03099084baff2be8fb/server/jobs/kube2store/slice-event-handler.go#L1 

```
	h.s.Update(func(tx *proxystore.Tx) {
		klog.V(3).Info("service ", service.Namespace, "/", service.Name, " topology key: ", svc.Spec.TopologyKeys)
		tx.SetService(service, svc.Spec.TopologyKeys)
		h.updateSync(proxystore.Services, tx)
	})
```

# Problem

I think these APIs start dissapearing in 1.22+, as i recently learned when updating the codebase to 1.23...  So we need to move to the new API to avoid compilation breaking for KPNG w/ 1.23 versions+

https://kubernetes.io/docs/concepts/services-networking/topology-aware-hints/

# Solution

So To solve this issue : 

- Look at how KPNG sends Topology hints to backend and document this
- Run KPNG and observe results of E2E Tests for Topology services
- Remove the above code that sets keys or replace w string[], confirm that theres some kind of regression.
- Now update to the 1.23 API, and re run these tests, observe the same regression
- Migrate the svc.Spec.TopologyKeys code to use the new Topology Hints API constructs, described in the blog post above.

This shouldnt be brutally difficult - it only touches the KPNG brain, i think... but I'm not 100% sure.  is a good issue for someone to own end to end that wants to get involved in KPNG and is looking for a medium difficulty challenge .

Since the  KPNG communitys bandwidth is a little limeited, it would i think be best to do this issue all in one, s o that we can review it in its holistic completion., as opposed to breaking it out into separate commits, especially since it involves breaking the old API  for how  kpng worked and requires looking at how it effects all backends --- and also --- i wonder if we need to make new upstream e2e tests ?

### LINKED ISSUE #317: Api discovery.k8s.io/v1beta1 is removed in K8s v1.25+
Replaced with discovery.k8s.io/v1. 

The `Topology` field used by kpng is removed from the API, probably replaced by [Hints](https://github.com/kubernetes/kubernetes/blob/2e64ae6d6253737448c7a7bf8a2f76989b13935c/staging/src/k8s.io/api/discovery/v1/types.go#L113).

There is a [DeprecatedTopology](https://github.com/kubernetes/kubernetes/blob/2e64ae6d6253737448c7a7bf8a2f76989b13935c/staging/src/k8s.io/api/discovery/v1/types.go#L101) also scheduled for removal.

Currently `kpng` can't be built on K8s `master`.

mcluseau:
@uablrek @astoycos my take on topology migration. It also includes zone filtering so we get `TopologyAwareHints` support here. Node name relies on the `EndpointSliceNodeName` feature gate, which is GA since 1.21.

Fixes #317 
Fixes #262


server/pkg/endpoints/node.go
```diff
@@ -48,17 +44,29 @@ func ForNode(tx *proxystore.Tx, si *localnetv1.ServiceInfo, nodeName string) (in
...
-
-		info.Endpoint.Local = epNodeName == nodeName
+		info.Endpoint.Local = info.Topology.Node == nodeName
 
 		if !info.Conditions.Ready {
 			return
 		}
 
+		if hints := info.Hints; hints != nil {
+			if len(hints.Zones) != 0 {
```
mcluseau, line 54: @robscott hi, can you confirm it's the spirit of the zones hint? :)

uablrek:
I applied this PR and applied it. Build works fine.

I was hoping that it would make `[Feature:ServiceInternalTrafficPolicy]` tests pass, but;
```
FAILED [sig-network] Services should respect internalTrafficPolicy=Local Pod (hostNetwork: true) to Pod [Feature:ServiceInternalTrafficPolicy]
FAILED [sig-network] Services should respect internalTrafficPolicy=Local Pod and Node, to Pod (hostNetwork: true) [Feature:ServiceInternalTrafficPolicy]
FAILED [sig-network] Services should respect internalTrafficPolicy=Local Pod to Pod [Feature:ServiceInternalTrafficPolicy]
```

mcluseau:
@uablrek I've added a test to make sure service traffic policy was correctly forwarded to the global service model, and it is. Combined with the test in `server/pkg/endpoints/node_test.go`, there's some good confidence that the chain is correct. But then, the backend shouldn't even know about the other endpoints. #287 comes with a more precise job on this. Maybe we should report this issue after all this is merged?

server/pkg/endpoints/node.go
robscott: I'm not that familiar with the code base here, but this looks reasonable in isolation. With that said, there's more logic here when it comes to handling the interactions with traffic policies - refer to the recently refactored upstream logic here: https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/topology.go#L40

mcluseau: thanks @robscott I'll open an issue to dig this

Issue #313, opened 2022-07-27
IPv6DualStack e2e tests fails for all backends


uablrek:
E2e tests executed with;
```
FOCUS='.*\[Feature:IPv6DualStack\]'
export SKIP='Disruptive|Serial|affinity|DNS|GCE|finalizer|ProxyTerminating'
```

### Test results

<details><summary>nft</summary>

```
PASSED [sig-network] [Feature:IPv6DualStack] should have ipv4 and ipv6 internal node ip
PASSED [sig-network] [Feature:IPv6DualStack] should create service with ipv6 cluster ip
PASSED [sig-network] [Feature:IPv6DualStack] should create service with ipv6,v4 cluster ip
PASSED [sig-network] [Feature:IPv6DualStack] should create a single stack service with cluster ip from primary service range
PASSED [sig-network] [Feature:IPv6DualStack] should create service with ipv4,v6 cluster ip
PASSED [sig-network] [Feature:IPv6DualStack] should create pod, add ipv6 and ipv4 ip to pod ips
PASSED [sig-network] [Feature:IPv6DualStack] should create service with ipv4 cluster ip
FAILED [sig-network] [Feature:IPv6DualStack] Granular Checks: Services Secondary IP Family [LinuxOnly] should function for node-Service: udp
FAILED [sig-network] [Feature:IPv6DualStack] Granular Checks: Services Secondary IP Family [LinuxOnly] should function for node-Service: http
PASSED [sig-network] [Feature:IPv6DualStack] should be able to reach pod on ipv4 and ipv6 ip
FAILED [sig-network] [Feature:IPv6DualStack] Granular Checks: Services Secondary IP Family [LinuxOnly] should function for pod-Service: http
FAILED [sig-network] [Feature:IPv6DualStack] Granular Checks: Services Secondary IP Family [LinuxOnly] should update endpoints: udp
FAILED [sig-network] [Feature:IPv6DualStack] Granular Checks: Services Secondary IP Family [LinuxOnly] should function for endpoint-Service: udp
FAILED [sig-network] [Feature:IPv6DualStack] Granular Checks: Services Secondary IP Family [LinuxOnly] should function for pod-Service: udp
PASSED [sig-network] [Feature:IPv6DualStack] Granular Checks: Services Secondary IP Family [LinuxOnly] should be able to handle large requests: udp
[...]

PR #319, opened 2022-07-31
test e2e: improve IPv6 handling and bump versions

mcluseau:
@uablrek I think it will impact #313 significantly.


mcluseau:
@nehaLohia27 @jayunit100 @qabpea does it looks good? To me it looks like IPv6 was just setup with default IP ranges and kpng not given the right CIDRs without those changes.

nehaLohia27 (approved):

qabpea:
That would explain a lot of the problems with the v6 tests.

--Per

Sent from Nine<http://www.9folders.com/>
________________________________

PR #320, opened 2022-08-01
watchstate: allow multiple stores per set

mcluseau:
Required to solve #314


mcluseau:
@gauravkghildiyal based on this PR, we can implement a solution to #314 this way: https://github.com/mcluseau/kpng/commit/0503286c2ed7da82fdf141fb5150a3e5b8ca16c3

gauravkghildiyal:
This looks great! Thanks for the quick resolution @mcluseau

PR #329, opened 2022-08-21
adding grpc SEND op_item logging/reasoning, package de-coupling from backend -> client

### LINKED ISSUE #327: [understandability] make nft/ and iptables/ packages more similar
Both NFT And IPTAbles import the localsink package, of course, but the import different structs from it.  To make it clear where their architecture is similar lets

- move the logic of iptables sink.go  into register.go OR
- make the register.go in nft separated out into sink.go for the part that imports the "sigs.k8s.io/kpng/client/localsink/fullstate"
- make it so the different Sink implementations are easy to reason about
- Add logging to the OpItem and Send functions (these run on the backends... but they are called from the brain)

jayunit100:
fixes #327 

- Documentation for localsink implementations
  - fullstate
  - decoder
  - filterreset
- Making import of fullstate explicit
- Document how cobra functions are added from Sinks
- Maybe some renaming so Sink's are easier to differentiate


astoycos:
I can take a stab here too if you need help @jayunit100 !

PR #334, opened 2022-08-24
Add gcc-multilib as dependency

vishnoianil:
eBPF backend needs this lib to compile the eBPF objects.

In absence of this library `go generate` will throw following error

`/usr/include/linux/types.h:5:10: fatal error: 'asm/types.h' file not
found`

Signed-off-by: Anil Kumar Vishnoi <vishnoianil@gmail.com>


astoycos:
Thanks @vishnoianil!

PR #335, opened 2022-08-29
Add replace for backends

papagalu:
go mod vendor will not pick up the code changes without the replace statements
@jayunit100  please also test to be 100% sure

### DIFF
--- a/cmd/go.mod
+++ b/cmd/go.mod
@@ -44,6 +44,13 @@ replace (
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.23.0
 	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.23.0
 	k8s.io/sample-controller => k8s.io/sample-controller v0.23.0
+	sigs.k8s.io/kpng/backends/ebpf => ../backends/ebpf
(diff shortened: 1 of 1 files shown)


astoycos:
Hrm... Shouldn't `go workspaces` take care of this? @mcluseau possibly you need to run `go work sync` to pickup the new changes

papagalu:
@astoycos 
im running go mod vendor, and it does not pick the backend changes, might be totally wrong tho

astoycos:
Right with [go  workspaces](https://go.dev/doc/tutorial/workspaces) you shouldn't need to manually run `go mod vendor` in the `/cmd` directory 

Are your changes not pulled in when running `./hack/test_e2e.sh -i ipv4 -b ebpf -d` or something of the sort? 

Because If so it'd be a build script issue most likely, not something with the go-mod

astoycos:
We got rid of these manual replace directives here https://github.com/kubernetes-sigs/kpng/commit/ece218f2a535bd88f7e3f738a948ccb31fc63334#diff-f1a968212f46fdc23571392dbada601265c55aec6360ae92e4020fcba94156b3L5

papagalu:
i believe this is not needed, thank you @astoycos !!!

PR #336, opened 2022-08-29
Fix mods From bringing in windows userspace

astoycos:
Fix mods after merging of windows userspace and add a make target for bumping in-tree K8s dependency versions (see https://github.com/kubernetes/kubernetes/issues/90358). Short of it is we need to ensure that dep versions are fixed based on the k8s version since technically k8s/k8s isn't meant to be vendored.

Also add windows userspace to the go workspace and remove many un-necessary `replace` directives in each backend


astoycos:
cc @jayunit100 and @mcluseau

astoycos:
Now there's a single replace directive in the window kernelspace go.mod and a few in the `cmd` go.mod but otherwise it should be a lot simpler

We should probably add CI to check if `go work sync` is happy before merging :)

astoycos:
/hold 

in favor of #333 :) 

Might still try to incorporate this make target

PR #337, opened 2022-08-30
WIP Build and run KPNG in Windows containers

marosset:
I'm working on running using KPNG instead of kube-proxy in a CAPZ cluster.
This PR adds 
- files needed to build a windows container image w/ kpng and files needed to start it
- deployment files for running kpng in a cluster on windows nodes.

I have this running on a CAPZ cluster!

In order to make more progress here several hard-coded values (--source-vip, --enable-dsr, --node-ip, --cluster-cidr) need to be exposed as flags for the winkernel backend.
I'll try and work on that as a separate PR.

Note - this is very much a WIP. We shouldn't be running `kube to-api` on every node but this was the easiest path forward for me :)

### DIFF
--- a/examples/windows/start-kpng.ps1
+++ b/examples/windows/start-kpng.ps1
@@ -11,12 +11,13 @@ while (-Not (Get-HnsNetwork | ? Name -EQ $NetworkName)) {
 }
 Write-Host "Found HNS network '$NetworkName'"
 
-# TODO: add node-ip and cluster-cidr here once those values are exposed as flags
+# TODO: and enable-dsr??
 $argList = @(`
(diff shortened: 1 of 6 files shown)


examples/windows/start-kpng.ps1
```diff
@@ -0,0 +1,38 @@
+$ErrorActionPrefernce = "Stop"
+
+Write-Host "Importing hns.pms1"
+Import-Module .\hns.psm1
+
+$NetworkName = "Calico"
```
jayunit100, line 6: make sure to disable kube-proxy if starting calico :)

PR #338, opened 2022-08-31
Remove static headerfiles

astoycos:
Remove static headerfiles and instead rely on 

UAPI from https://github.com/libbpf/libbpf/tree/master/include/uapi/linux
bpf helpers from https://github.com/libbpf/libbpf/tree/master/src

from libbpf version v0.8.0 

Add new make target which handles downloading these headers if needed


astoycos:
/hold

I want to get some more opinions before merging this

backends/ebpf/bpf/cgroup_connect4.c
```diff
@@ -1,14 +1,15 @@
 /* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
 /* Copyright Authors of Cilium */
-#include <linux/bpf.h>
-#include <linux/in.h>
+#include "include/uapi/linux/bpf.h"
+#include "src/bpf_helpers.h"
```
dave-tucker, line 4: What we use in `bpfd` is this:

https://github.com/redhat-et/bpfd/blob/8d837c1acb9a62018d904d1240e37376ce2e1e73/xtask/src/build_ebpf.rs#L103

Effectively, from the `libbpf` checkout `make INCLUDEDIR="./include/bpf" install_headers`.
If you invoke clang with `-I/path/to/libbpf/include' your includes then become:
```
#include <uapi/linux/bpf.h>
#include <bpf/bpf_helpers.h>
```
... which sits better with me.

astoycos: ACK thanks for the review, done

dougsland (approved):

PR #343, opened 2022-09-06
Bumped e2e CI to k8s latest release

nerdalert:
Signed-off-by: Brent Salisbury <bsalisbu@redhat.com>


nerdalert:
- Moving to draft until I go through failing tests to see if this creates any new ones since we currently fail open in CI mode. 
- On the fail open e2e topic, it might be worth expanding the skip list and fail CI on new test failures going forward along with working to rectify failing skips. I will open an issue on that and look into it.

Ty!

PR #346, opened 2022-09-16
Add info about allocateNodePorts=false to the service-proxy docs

tssurya:
Clarifies that setting allocateNodePorts=false to a load balancer service has nothing to do with the healthcheckNodePorts.
/cc @danwinship PTAL


danwinship:
/lgtm
/approve
except I don't think I can actually approve here

PR #348, opened 2022-09-20
Update kpng-local-up.sh and fix potential IPVS panic

aryan9600:
While trying to run the project locally using `hack/kpng-local-up.sh`, I encountered several problems:
* Using the ipvs backend used to cause a panic, because the base distro being used in the kind node was out of date hence there was no `proc/sys/net/ipv4/vs` present, resulting in `Setup()` to fail silently. Later when a new service gets synced, `AddIP()` panics because `s.dummy` never got set.
* The `hack/kpng-local-up.sh` is out of date using `envsubst` instead of `j2`

### DIFF
--- a/hack/kpng-local-up.sh
+++ b/hack/kpng-local-up.sh
@@ -14,22 +14,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+source ./utils.sh
 
 # build the kpng image...
-# TODO Replace with 1.22 once we address 
-#: ${KIND:="kindest/node:v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6"}
-: ${KIND:="kindest/node:v1.22.0@sha256:b8bda84bb3a190e6e028b1760d277454a72267a5454b57db34437c34a588d047"}
+: ${KIND:="kindest/node:v1.22.13@sha256:4904eda4d6e64b402169797805b8ec01f50133960ad6c19af45173a27eadf959"}
 : ${IMAGE:="gauravkghildiyal/kpng:latest"}
 : ${PULL:="IfNotPresent"}
-: ${E2E_BACKEND:="nft"}
+: ${BACKEND:="nft"}
 : ${CONFIG_MAP_NAME:=kpng}
 : ${SERVICE_ACCOUNT_NAME:=kpng}
 : ${NAMESPACE:=kube-system}
 : ${KPNG_DEBUG_LEVEL:=4}
+: ${BACKEND_ARGS:="['local', '--api=unix:///k8s/proxy.sock', 'to-${BACKEND}', '--v=${KPNG_DEBUG_LEVEL}']"}
 
-export IMAGE PULL E2E_BACKEND CONFIG_MAP_NAME SERVICE_ACCOUNT_NAME NAMESPACE KPNG_DEBUG_LEVEL
-
-echo -n "this will deploy kpng with docker image $IMAGE, pull policy '$PULL' and the '$E2E_BACKEND' backend. Press enter to confirm, C-c to cancel"
+echo -n "this will deploy kpng with docker image $IMAGE, pull policy '$PULL' and the '$BACKEND' backend. Press enter to confirm, C-c to cancel"
 read
 
 function build_kpng {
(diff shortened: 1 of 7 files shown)


hack/utils.sh
```diff
@@ -0,0 +1,121 @@
...
+    [ -d "${directory}" ]
+    if_error_exit "Directory \"${directory}\" does not exist"
+
+    case ":${PATH:-}:" in
+        *:${directory}:*) ;;
+        *) PATH="${directory}${PATH:+:$PATH}" ;;
+    esac
+}
+
+command_exists() {
```
astoycos, line 75: Can you keep the comment here :)

aryan9600: sorry it got lost somehow while refactoring, i added it back :)

hack/kpng-local-up.sh
```diff
@@ -72,8 +70,18 @@ function install_k8s {
...
-    envsubst <kpng-deployment-ds.yaml.tmpl >kpng-deployment-ds.yaml
+    # Setting vars for generate the kpng deployment based on template
+    kpng_image="${IMAGE}" \
+    image_pull_policy="${PULL}" \
+    backend="${BACKEND}" \
+    config_map_name="${CONFIG_MAP_NAME}" \
+    service_account_name="${SERVICE_ACCOUNT_NAME}" \
+    namespace="${NAMESPACE}" \
+    e2e_backend_args="${BACKEND_ARGS}"\
+    j2 ./kpng-deployment-ds.yaml.j2 -o ./kpng-deployment-ds.yaml
```
astoycos, line 83: So I think @qabpea Just removed this? (I had added it originally) I'd like him to take a look as well

astoycos: I'm alright with just using j2 for now though!

astoycos:
Thanks for doing this! I just want to see what Per has to say about re-using J2 since he just merged https://github.com/kubernetes-sigs/kpng/pull/345

astoycos:
@aryan9600 Overall this looks great and I tested locally, do you mind also adding a quick note to the readme in `/hack/` Explaining how to spin up a devel cluster with `./hack/test_e2e.sh  -d -i ipv4 -b <backend>` and the difference between that and `./hack/kpng-local-up.sh`? 

I'll go ahead and give it an lgtm ig you want to tackle that in a follow up PR

/lgtm

hack/kpng-local-up.sh
qabpea: Is the question where bgp2go is installed?
I did this
    echo "'bpf2go' not found, installing with 'go'"
	# set GOBIN to bin_directory to endure that binary is in search path
	export GOBIN=${install_directory}
	
	#remove GOPATH just to be sure
	export GOPATH=""


The problem I was solving was when a user do not have a GO environment setup, but go is installed and bgp2go is not in the path
go install  installs in the default GOBIN directory which is often not part of the users path.
GOBIN is therfor  set to the local bin dir.

But this should never be triggered unless there is no bgp2go in the path to begin with.

astoycos:
Hrm 
```
# github.com/cilium/ebpf/internal
../../../go/pkg/mod/github.com/cilium/ebpf@v0.9.3-0.20220920164205-03b648ce0c4c/internal/deque.go:6:14: syntax error: unexpected any, expecting ]
../../../go/pkg/mod/github.com/cilium/ebpf@v0.9.3-0.20220920164205-03b648ce0c4c/internal/deque.go:12:16: syntax error: unexpected [, expecting comma or )
../../../go/pkg/mod/github.com/cilium/ebpf@v0.9.3-0.20220920164205-03b648ce0c4c/internal/deque.go:16:16: syntax error: unexpected [, expecting comma or )
../../../go/pkg/mod/github.com/cilium/ebpf@v0.9.3-0.20220920164205-03b648ce0c4c/internal/deque.go:21:16: syntax error: unexpected [, expecting comma or )
../../../go/pkg/mod/github.com/cilium/ebpf@v0.9.3-0.20220920164205-03b648ce0c4c/internal/deque.go:37:16: syntax error: unexpected [, expecting comma or )
../../../go/pkg/mod/github.com/cilium/ebpf@v0.9.3-0.20220920164205-03b648ce0c4c/internal/deque.go:40:2: syntax error: non-declaration statement outside function body
../../../go/pkg/mod/github.com/cilium/ebpf@v0.9.3-0.20220920164205-03b648ce0c4c/internal/deque.go:52:16: syntax error: unexpected [, expecting comma or )
../../../go/pkg/mod/github.com/cilium/ebpf@v0.9.3-0.20220920164205-03b648ce0c4c/internal/deque.go:53:6: zero redeclared in this block
	/home/runner/go/pkg/mod/github.com/cilium/ebpf@v0.9.3-0.20220920164205-03b648ce0c4c/internal/deque.go:38:6: previous declaration
../../../go/pkg/mod/github.com/cilium/ebpf@v0.9.3-0.20220920164205-03b648ce0c4c/internal/deque.go:55:2: syntax error: non-declaration statement outside function body
../../../go/pkg/mod/github.com/cilium/ebpf@v0.9.3-0.20220920164205-03b648ce0c4c/internal/deque.go:70:16: syntax error: unexpected [, expecting comma or )
../../../go/pkg/mod/github.com/cilium/ebpf@v0.9.3-0.20220920164205-03b648ce0c4c/internal/deque.go:53:6: too many errors
note: module requires Go 1.[18](https://github.com/kubernetes-sigs/kpng/actions/runs/3092361770/jobs/5003547330#step:3:19)
```

astoycos:
/hold 

Just until  CI is green

aryan9600:
@astoycos  I pinned `cillium/ebpf` to v0.9.2 in `hack/test_e2e.sh`, to avoid this, i'll do it in the github workflow as well

hack/kpng-local-up.sh
astoycos: It looks like we're not changing that behavior here :)

astoycos:
@jayunit100 This should be good :)

PR #357, opened 2022-09-22
fix #353: remove legacy endpoints support

### LINKED ISSUE #353: remove the endpoints logic in favor of only using endpoint slices
right now the kube2store has a fair amount of logic for setting up (if/else) endpoint controllers, and also duplicate mapping logic to convert the incoming k8s api endpoints objects to kpng objects.

if we just ditch that logic (since , endpoints now deprecated https://kubernetes.io/docs/reference/using-api/deprecation-guide/#endpointslice-v125 ).... we can clean some tech debt

also this is a  easy one for someone familiary w/ go , so tagging as gfi 

/good-first-issue

mcluseau:
Sorry @jayunit100 , I didn't see the "good first issue" :smiling_face_with_tear: I got too motivated by the cleanup


heisaman:
so motivated, leaving us newcomers amazed...

jayunit100:
haha thank you sir !
/lgtm
/approve

PR #360, opened 2022-10-04
Fix hack README.md run from source

loop0:
This PR fixes the documentation on the Run from source section which repeats the env var IMAGE with the invalid reference format and also fixes the path to run the `kpng-local-up.sh` script


knabben:
thanks

/lgtm
/approve

PR #363, opened 2022-10-11
server: move proxystore up one level

mcluseau:
@astoycos move proxystore up one level so it's more obvious that it's the core thing of kpng's server side.


astoycos:
Sweet I think this is nice for readability :)

server/jobs/api2localdiff/api2localdiff.go
```diff
@@ -1,17 +0,0 @@
-/*
```
astoycos, line 1: Why Was this here in the first place? It seems like a place to implement API -> Local Streaming AND diffing?

astoycos:
/lgtm 

Thanks for this bit!

server/jobs/api2localdiff/api2localdiff.go
mcluseau: an old (1y+) mistake I think, never had any kind of code

PR #364, opened 2022-10-11
server: add some doc with pointers to each function of the server

mcluseau:
@astoycos some doc to help

Depends (softly) on #363 for the proxystore's links.

### DIFF
--- a/server/doc.go
+++ b/server/doc.go
@@ -1,0 +1,34 @@
+// Package server contains every kpng's function related to serving its API.
+//
+// The general approach is that you want to provide a [proxystore.Store] and make something of it.
+// This store contains what we call the global model, representing the cluster-wide information
+// needed by kpng to provide what we call the local model, representing the node-level information.
+//
+// There are 3 ways to provide a [proxystore.Store], all using jobs:
+//   - [jobs/kube2store]: from a k8s' apiserver;
+//   - [jobs/api2store]: from another kpng instance serving the global API;
+//   - [jobs/file2store]: from a serialized global model.
+//
+// From a [proxystore.Store], you can:
+//   - serve it over the kpng's API ([jobs/store2api]);
+//   - serialize the global model to a file ([jobs/store2file]);
+//   - stream the changes to the global model ([jobs/store2globaldiff]);
+//   - stream the changes to the local model ([jobs/store2localdiff]).
+//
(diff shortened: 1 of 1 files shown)


server/doc.go
```diff
@@ -0,0 +1,34 @@
...
+//
+// There are 3 ways to provide a [proxystore.Store], all using jobs:
+//   - [jobs/kube2store]: from a k8s' apiserver;
+//   - [jobs/api2store]: from another kpng instance serving the global API;
+//   - [jobs/file2store]: from a serialized global model.
+//
+// From a [proxystore.Store], you can:
+//   - serve it over the kpng's API ([jobs/store2api]);
+//   - serialize the global model to a file ([jobs/store2file]);
+//   - stream the changes to the global model ([jobs/store2globaldiff]);
```
astoycos, line 15: Why would we want to 
1. Store a global model in a proxy store AND 
2. Stream that global model? 

Is that to facilitate the flow where we don't convert to local until TORX?

```
# central # kpng kube to-api --server central-kpng:1234 -----> Serves Global API 
# DCx     # kpng api --api central-kpng:1234 to-api --server dcx-kpng:1234 --------> Serves Global API 
# ToRx    # kpng api --api dcx-kpng:1234     to-api --server torx-kpng:1234 ----------> Serves Global API 
# node    # kpng local --api torx-kpng:1234  to-nft --... ------> Requests Local API
```

server/doc.go
```diff
@@ -0,0 +1,34 @@
...
+//   - stream the changes to the local model ([jobs/store2localdiff]).
```
astoycos, line 16: Maybe we should add a note on the `watchstate` package here? It's pretty crucial for the streaming cases right? 

Something like:

The WatchState Package is what we use internally to actually facility the streaming of Diffs to the backend Sinks?

server/doc.go
astoycos: Cool yeah I need to add this to the Readme I think the three main deployments I'll show is 

1. Single node one process 

2. Split deploymnt One Tier  (i.e just Central -> Node) 

3. Split deployment Multiple Tiers (i.e what I described above) 

Is there anything else I'm missing there?

server/doc.go
mcluseau: I didn't go inside pkg in this iteration. We can add more later, esp. you as you're going through it :)

PR #365, opened 2022-10-11
Improve userspace backend 

aryan9600:
* fix endpoints deletion handling in userspace: fixes https://github.com/kubernetes/kubernetes/blob/fbb6ccc0c62d2431dc3a18a4130d7fbae5c05acd/test/e2e/network/service.go#L792
* fix port to endpoint mapping in userspace: fixes  https://github.com/kubernetes/kubernetes/blob/fbb6ccc0c62d2431dc3a18a4130d7fbae5c05acd/test/e2e/network/service.go#L792

### DIFF
--- a/backends/userspacelin/roundrobin.go
+++ b/backends/userspacelin/roundrobin.go
@@ -307,18 +307,6 @@ func (lb *LoadBalancerRR) OnEndpointsAdd(ep *localv1.Endpoint, svc *localv1.Serv
 // 	}
 // }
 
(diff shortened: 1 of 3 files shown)


jayunit100:
/approve
/lgtm

btw can you add some comments to that Userspace backend struct (i.e. udpIdleTimeout and all those fields, make sure they're better described) maybe in a follow on thanks @aryan9600 !!!!!!!!!

PR #388, opened 2022-10-18
Update GRPC API naming

astoycos:
Move Endpoints Service to Local Service.

Move Endpoints/Services_set to
Local(Endpoints/Services)_set.

fixes https://github.com/kubernetes-sigs/kpng/issues/387

PR #390, opened 2022-10-31
bifurcate GRPC  data model into local and global  implementations

### LINKED ISSUE #391: separating kpng cord and kpng backend APIs
(just did this at kpngcon)  details are in the PR

#390

jayunit100:
fixes #391 

At KPNGCON we went through the GRPC and split into two separate
data models: one for local, and one for global state.

This makes it conceptually easier for use to reason about info that 
is sent to individual nodes. vs info that is global WRT the overall networking statespace.

As an example, 

A global object would be something like a topology... Something that isn't relevant to an individual node, whose job is simply to manage routing rules, and is more a "business object" of the overall transformation of the K8s API object to proxy routing rules. 

Local objects are (Services, IPs, protocols, Watch details, and so on).  Since really they relate a local KPNG backend to the server (for example, a "Watch" object only makes sense from a local backend perspective... it defines a node, sends it to the server (brain), and then receives events)


jayunit100:
@astoycos is this good w/ you ?

api/globalv1/api.proto
```diff
@@ -0,0 +1,77 @@
...
+
+import "api/localv1/api.proto";
+
+option go_package = "sigs.k8s.io/kpng/api/globalv1";
+
+//
+// global (cluster-wide) model
+//
+
+message ServiceInfo {
```
astoycos, line 27: Kind of a Nit but I think we should add comments to all of these types just like we do in K8s, It might be a bit tedious but may be helpful for folks trying to understand it 

(we also could tackle it in a follow up PR)

api/globalv1/api.proto
```diff
@@ -0,0 +1,77 @@
...
+  localv1.Service Service = 2;
+}
+
+message EndpointInfo {
+  uint64 Hash = 1;
+
+  string Namespace = 2;
+  string SourceName = 3;
+  string ServiceName = 4;
+  reserved 5;
```
astoycos, line 39: what does `reserved` signify here?

api/globalv1/api.proto
```diff
@@ -0,0 +1,77 @@
...
+  string PodName = 8;
+
+  localv1.Endpoint Endpoint = 6;
+  EndpointConditions Conditions = 7;
+  TopologyInfo Topology = 9;
+  TopologyHints Hints = 10;
+}
+
+message EndpointConditions {
```
astoycos, line 48: Can we get rid of the indirection if `Ready` is the only condition we care about?

api/localv1/api.proto
```diff
@@ -14,20 +14,33 @@
...
-option go_package = "sigs.k8s.io/kpng/server/pkg/api/localnetv1";
+option go_package = "sigs.k8s.io/kpng/api/localv1";
 
-service Endpoints { // FIXME rename to Local
+service Sets {
     // Returns all the endpoints for this node.
     rpc Watch (stream WatchReq) returns (stream OpItem);
 }
 
+// any time were watching for the local model, we need to send the
```
astoycos, line 26:
```suggestion
// To request ENLS(Expected Node Local State) a client must specify the desired NodeName
```

api/localv1/api.proto
```diff
@@ -14,20 +14,33 @@
...
     string NodeName = 1;
 }
+enum Set {
+    UnknownSet = 0;
+
+
+    ServicesSet = 1;
+    EndpointsSet = 2;
 
+    // FIXME move to a 3rd generic proto ???
```
astoycos, line 39: Hrm can't this just live in `globalv1`?

astoycos:
So I think this general change is great and helps with readability/separation, my only worry is that having two separate APIs _could_ make things a bit more difficult to maintain in the future.  It's either we go with this OR just stick to one API and update the naming to better separate the global/local model (like I attempted here https://github.com/kubernetes-sigs/kpng/pull/388) 

>A global object would be something like a topology... Something that isn't relevant to an individual node, whose job is simply to manage routing rules, and is more a "business object" of the overall transformation of the K8s API object to proxy routing rules.

But wait we can derive the local state from this global state, so really it's just a super set of ENLS(expected node local state right? 

>Local objects are (Services, IPs, protocols, Watch details, and so on). Since really they relate a local KPNG backend to the server (for example, a "Watch" object only makes sense from a local backend perspective... it defines a node, sends it to the server (brain), and then receives events)

I think it would go a long way if we were super detailed on what information exactly is defined in the global state vs local state.... maybe in the comments maybe in another doc

api/localv1/api.proto
jayunit100: It can it it does but... these global fields aren't referenced in the local data model.  hence the. fixme. I f nothing else, it helps w/ long term understanding of what we were thinking when we included these unused fields in the local proto

api/globalv1/api.proto
jayunit100: agree we need to comment them all .  actually this entire pr stems from the fact that me and amim couldn't figure out what any of these structs where differentiated for .... so we got mikael to split up the files. 

lets do that in a follow

jayunit100:
@astoycos we went back and forth on "Namespacing in the single proto" vs namespacing by separate protos.  I agree it makes no difference but mikael and amim both leaned towards the 2nd proto w package contexts over Java style explicit variable names

astoycos:
/lgtm
/approve

I'll go with this for now :)

astoycos:
closes #388

PR #392, opened 2022-11-03
Updated test_e2e to use K8s v1.25.3 and kind 0.17.0, plus small bug fix

qabpea:
Updated k8s and kind verisions plus minor fixes to make the script work using -n flag again


hack/test_e2e.sh
```diff
@@ -621,7 +623,7 @@ function run_tests {
    export KUBE_CONTAINER_RUNTIME_ENDPOINT=unix:///run/containerd/containerd.sock
    export KUBE_CONTAINER_RUNTIME_NAME=containerd
 
-   ${e2e_dir}/bin/ginkgo --nodes="${GINKGO_NUMBER_OF_NODES}" \
```
astoycos, line 624: @qabpea why not just make this `"${bin_dir}/ginkgo"` instead of adding new args to run_tests and `create_infrastructure_and_run_tests`?  

Even better what if we just passed `${bin_dir}` to `run_tests()` and then that function can handle calling `gingko` and `e2e.test` from there

We keep adding arguments and things are getting confusing IMO

hack/test_e2e.sh
```diff
@@ -872,24 +874,26 @@ function create_infrastructure_and_run_tests {
...
-    #   arg6: developer_mode                                                  #
-    #   arg7: <ci_mode>                                                       #
-    #   arg8: deployment_model                                                #
-    #   arg9: export_metrics                                                  #
+    #   arg5: ginkgo                                                          #
+    #   arg6: suffix                                                          #
+    #   arg7: developer_mode                                                  #
+    #   arg8: <ci_mode>                                                       #
+    #   arg9: deployment_model                                                #
+    #   arg10: export_metrics                                                  #
```
astoycos, line 882: SUPER NIT: :) That last `#` is pushed a space too far

hack/test_e2e.sh
```diff
@@ -1236,7 +1245,7 @@ fi
 if [[ -n "${ip_family}" && -n "${backend}" ]]; then
     main "${ip_family}" "${backend}" "${ci_mode}" "${e2e_dir}" "${bin_dir}" "${dockerfile}" \
          "${suffix}" "${cluster_count}" "${erase_clusters}" "${print_report}" "${devel_mode}" \
-         "${deployment_model}" "${run_tests_on_existing_cluster}" "${export_metrics}"
+         "${deployment_model}" "${run_tests_on_existing_cluster}" "${export_metrics}"
```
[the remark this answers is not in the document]
astoycos:
```suggestion
         "${deployment_model}" "${run_tests_on_existing_cluster}" "${export_metrics}" 
```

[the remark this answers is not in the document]
qabpea: not sure either, extra spaces at the end? I removed any extra spaces

qabpea:
One small change
  export KUBE_CONTAINER_RUNTIME=remote
   export KUBE_CONTAINER_RUNTIME_ENDPOINT=unix:///run/containerd/containerd.sock
   export KUBE_CONTAINER_RUNTIME_NAME=containerd

   ${ginkgo} --nodes="${GINKGO_NUMBER_OF_NODES}" \

The “${e2e_dir}/bin/ginkgo” does not work if the -B option is used, it is replaced with    ${ginkgo} now.

//Per

hack/test_e2e.sh
```diff
@@ -583,15 +583,17 @@ function run_tests {
      # Arguments:                                                              #
      #   arg1: e2e directory                                                   #
      #   arg2: e2e_test, path to test binary                                   #
-     #   arg3: parallel ginkgo tests boolean                                   #
+     #   arg3: ginkgo path to ginko binary                                     #
```
astoycos, line 586: I was basically saying just remove this `arg3` and `arg4` from the function and replace it with `${bin_dir}`

qabpea: In that case, we should change arg 2 as well, the path to e2e_test is ${bin_dir}/e2e.test.

I also feel like it time to change the use of "kind" directly and trust the path to ${bin_dir}/kind

astoycos: I'm find with that, we never configure those binaries to be "different" anyways

qabpea: I changed to use ${bin_dir} everywhere needed and also made sure that the binaries in ${bin_dir} is always access through ${bin_dir}/command

astoycos:
/lgtm

Thanks @qabpea!

Issue #352, opened 2022-09-20
  e ) erase_clusters=false ;;... doesnt work


jayunit100:
hey yall im trying to 

- hack up the test_e2e.sh script to do `erase_clusters=false` but im not seeing any effect...

```

while getopts "i:b:B:cdD:eE:n:ps:" flag
do
    case "${flag}" in
        i ) ip_family="${OPTARG}" ;;
        b ) backend="${OPTARG}" ;;
        c ) ci_mode=true ;;
        d ) devel_mode=true ;;
        e ) erase_clusters=false ;;
```

but then if i run 

```
eval "$(curl -sL https://raw.githubusercontent.com/travis-ci/gimme/master/gimme | GIMME_GO_VERSION=1.18beta1 bash)" go version go1.18beta1 linux/amd64
./test_e2e.sh -i ipv4 -b iptables 


```

I ultimately end up with cluster gettin deleteddd :( :( :( 

```
+==================================================================+
                Erasing kind clusters
+==================================================================+
[ PASSED ] Cluster kpng-e2e-ipv4-iptables deleted.
```

/good-first-issue

emabota:
Hi @jayunit100, 
I am a New Contributor and would like to work on this issue. Can you please provide some orientation on how to approach this issue? Thanks 
/assign

jayunit100:
Hi just try to run the hack/test_e2e script w iptables option and use the -e flag you will see what I mean :)

emabota:
Hi @jayunit100, 
while trying to replicate this issue, whenever I run the script `./test_e2e.sh -i ipv4 -b iptables` it fails with the following end message: 

> ERROR: unknown cluster "[ \x1b[34mINFO\x1b[0m ] kpng-e2e-ipv4-iptables" 

Can you, please, help me figure out why it's happening? Find attached the "log" of the script.
[test_e2e_status.log](https://github.com/kubernetes-sigs/kpng/files/10114534/test_e2e_status.log)

jayunit100:
Ouch looks like you didn't get a cluster is kind installed ? Any other suspicious logs ?

emabota:
This was fixed here #399 
> Hi @jayunit100, while trying to replicate this issue, whenever I run the script `./test_e2e.sh -i ipv4 -b iptables` it fails with the following end message:
> 
> > ERROR: unknown cluster "[ \x1b[34mINFO\x1b[0m ] kpng-e2e-ipv4-iptables"
> 
[...]

PR #399, opened 2022-12-02
using echo instead of info_message to persist cluster_name

daman1807:
### Problem

hack/test_e2e.sh#L641 uses **info_message** function to persist cluster name to file, which adds _ANSI color escape sequences and INFO label_ along with cluster name.

On disk clustername becomes **[ \x1b[34mINFO\x1b[0m ] kpng-e2e-ipv4-iptables** instead of **kpng-e2e-ipv4-iptables**

### Fix
Using **echo** to persist cluster name instead of **info_message**

### DIFF
--- a/hack/test_e2e.sh
+++ b/hack/test_e2e.sh
@@ -638,7 +638,7 @@ function create_infrastructure_and_run_tests {
     create_cluster "${cluster_name}" "${bin_dir}" "${ip_family}" "${artifacts_directory}" "${ci_mode}"
     wait_until_cluster_is_ready "${cluster_name}" "${bin_dir}" "${ci_mode}"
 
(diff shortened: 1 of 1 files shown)


astoycos:
Thanks for the PR @daman1807 

@jaypipes Can you approve the workflow here in the GUI so the actions/ CI run? (don't just /approve the PR that won't work)

astoycos:
@daman1807 Did this break things when running locally? And if so do you have any idea why our CI didn't pick it up?

daman1807:
@jayunit100 this also fixes #352

jayunit100:
We should get rid of the colors in general imo

jayunit100:
Thanks for this!!! Can u link to the issue this fixes ?

daman1807:
this fixes #402

Issue #400, opened 2022-12-03
e2e test suite fails but job succeeds


aojea:
```
Summarizing 2 Failures:
  [FAIL] [sig-network] Services [It] should be updated after adding or deleting ports 
  test/e2e/network/service.go:966
  [FAIL] [sig-network] Services [It] should serve multiport endpoints from pods  [Conformance]
  test/e2e/network/service.go:922

Ran 306 of 7067 Specs in 666.762 seconds
FAIL! -- 304 Passed | 2 Failed | 0 Pending | 6761 Skipped


Ginkgo ran 1 suite in 11m11.265497035s

Test Suite Failed
```

However, the job is green https://github.com/kubernetes-sigs/kpng/actions/runs/3584852487/jobs/6032067673

???

jayunit100:
ahhh  so, yes, we want to "pass" because its not a regression, but i see what you mean . 

Should we add a separate (non-blocking) test that runs these 2 known failure cases ? or just make it fail every time ?

aojea:
those tests are exercising common functionality of Services, not having them passing means that kpng is not complaint, if you really want to offer kpng to users those test have to be passing

PR #401, opened 2022-12-03
Fail job if test suite fails

aojea:
Fixes: https://github.com/kubernetes-sigs/kpng/issues/400


jayunit100:
after this lets then separate tests "sig network stable " tests from the "sig network kpng failing" tests so that its easy for us to know about regression

jayunit100:
ha now we have to make it so that the bot merges it even though the tests fail i guess

aojea:
You can skip and add Todo

qabpea:
This has been fixed in PR #437, This PR can now be closed

aojea:
Thanks Per

PR #411, opened 2022-12-22
Making K2S explicit and more readable

### LINKED ISSUE #412: Remove stamp coupling and lambdas and make a2s, k2s, ... easy to trace
Right now the cmd/ package is tricky to read:

- We have some unnecessary types
- We have some unnecessary files (ip-port.go, maybe others?) that can be isolated to the minimal places.
- We create a lambda
- The lambda implements a type 
- The function type has methods on it  that are callable
- We then call the type w/ various toAPI, toLocal, and so on methods

The problem here is that we use types, rather than inputs, to configure these commands, this means that its not clear when reading the code *what* the inputs to these are (unless you go read the type, which is simply called something like "setupFunc", and which is later referenced as "c()" inside of these calls.

Since the indirection here puts a cognitive load on folks wanting to build their own Client impl,  We should:

- Comment exactly and name the Cobra generators explicitly
- Decouple the inputs from the toAPI, toLocal, and so on from any internal Types, and from each other
- Decouple the k2s/a2s/... from the underlying store builder as much as possible (i.e. remove wrapper types)

### LINKED ISSUE #414: ParallelExample tests can flake
the 1 3 8 millisecond waits arent enough for GH Actions.

jayunit100:
fixes #412 
fixes #414 also


client/plugins/ip-port.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package main
+package conntrack
```
knabben, line 17: should this be on `conntrack` folder, otherwise this is on plugins module.

jayunit100: yeah, removing we dont need that file good catch thanks amim

jayunit100:
dont lgtm yet

PR #415, opened 2022-12-23
Finish adding lambdas explicitly and also fix some kpng-local-up.sh errors related to legacy mode

jayunit100:
Was missing the lambda internalization from last commit , second half of #411 , fixes the first half of #416


jayunit100:
was missing a commit in my last MR, added here and also fixed kpng-local-up.sh to use kind , and not calico, which gets us (part of) the way there, towards fixing the kpng-local-up.sh issue

PR #417, opened 2022-12-31
IPVS Fullstate Implementation

daman1807:
# IPVS Fullstate Implementation

This implementation simply requests the creation of all the network objects required to program Kubernetes services data path for every callback, leverages diffstore to maintain the states of all the network components, and operates only on the changes returned by the diffstore.

## 1. Controller
Controller on initialization set sysctls for IPVS, creates dummy interface, initializes IPSets, writes required IPTable rules and most importantly handles the full state callback.
On processing full state controller interacts with **Managers** which acts as a proxy to the actual IPVS and IPSet resource creation.

- **addServiceEndpointsForClusterIP**
  The logic for programing a ClusterIP service
- **addServiceEndpointsForNodePort**
  The logic for programing a NodePort service
- **addServiceEndpointsForLoadBalancer**
  The logic for programing a LoadBalancer service
## 2. Managers
Manager leverages diffstore for storing all the resource manipulation operations (create virtual server, add destination, add entry to ipset) required to render the full state and only acts on the changes in the store.
- **ipvs**

  Resource definitions and methods for IPVS manipulation.
  IPVS manager holds virtual server and destination definitions
- **ipsets**

  Resource definitions and methods for IPSets manipulation.

  IPSets manager
- **iptables**

  Resource definitions and methods for IPTables manipulation.


mcluseau:
tldr: I'm worried about the NIH / duplication of pre-existing logic in kpng that's supposed to help backend writers. Be it unknown to the implementer even after pairing with @jayunit100 or not understood, I consider it as my failure by giving a model that's too split. As discussed with @astoycos a few weeks ago, it would be easier to just get rid of the sink layer and just push fullstate as the only way, and from here see if/where it fails at scale.

Examples:
- `gen.go`: it looks a lot like `kpng-backend-torture`
- storing the local model in the backend: it *is* what the fullstate sink does
- `patch.go` & related: it is what diffstore provides *if* you store your business objects in it instead of the local model

I'll take a deeper look now and probably send a PR to @daman1807's fork.

mcluseau:
@daman1807 take a look here https://github.com/mcluseau/kpng/blob/ipvs-fullstate/backends/ipvsfullstate/ipsets/ipsets.go as it is exactly what I wanted to show you.

The idea here is that you just push set entries in (using `AddEntry`). After processing the `Callback`'s input, you call `Done()`, get and apply the changes using `Created()` and `Deleted()` (giving you ipsets and entries). Since there's a logic of flushing sets, I allowed myself to do the fancy filtering of deleted entries at this level too.

The thing is really to thing about the actual system/low-level objects you want to manage, and use diffstore to know changes should be applied. The same idea can be seen in the nft backend here https://github.com/mcluseau/kpng/blob/ipvs-fullstate/backends/nft/nftable.go#L42

daman1807:
> The thing is really to thing about the actual system/low-level objects you want to manage, and use diffstore to know changes should be applied. The same idea can be seen in the nft backend here https://github.com/mcluseau/kpng/blob/ipvs-fullstate/backends/nft/nftable.go#L42

TBH, I realized this during my implementation and discussed it with @jayunit100 too.  We kept this thing for the second iteration. Should I change it in this iteration only?

mcluseau:
@daman1807 yes, the way with fullstate is 2 phases: (1) translate the local to system object definitiions (iterating other the given chan) and (2) apply changes to those catched by the diffstores, in the way that makes sense (what contains what, order constraints, etc).

If I may, I'd prefer to avoid doing this change in a second iteration for this 3rd (4th?) IPVS implementation.

daman1807:
@mcluseau got it. I will update my implementation then to adapt this behaviour.

jayunit100:
sgtm, I have no opinion on any of this :) as long as the actual IPVS tests *pass* and we use fullstate thats all i care about :) .  So... if we do the local model to system object definition thing here, or later, thats up to @daman1807 and @mcluseau !

jayunit100:
overall though, yeah, lets kill the sinks.  the indirection makes it super complicated  .  I dont think that should impact *this PR* though

mcluseau:
@jayunit100 killing the sink is definitely not the subject of this PR yes, it will take a bit of time because some (important) backends are using non-fullstate sinks (ie iptables).

mcluseau:
I agree, iptables is translated to nftables anyway (in the iptables tools or maybe even in the kernel, I didn't check), and using the nftables formalism (of the `nft` tool) gives a better control and more understandable rules.

astoycos:
So Let me preface this with the fact that I am not at all an nft or ipvs expert :) however, 

> Now, store.Updated() will return [ {A [A1 A2 A3]}  {B [B1 B2]} ] but for IPVS implementation we are only interested in A3 and B3. So I created two diffstores, one for services and one for endpoints, and created a structure ServiceEndpointsPatch to hold the difference, and associate related services and endpoints back together.
> In this example there will be two ServiceEndpointsPatchs [ {A [A3+]} ] and [ { B {B3-]} ].
Hiya @daman1807 This is great work, thanks so much for the documentation :) 

At a first glance it seems to me like data is flowing like so... 

kube api (incremental add/update/delete events) -> Proxystore (keeping fullstate) -> Watchstate (sending a partial diff) -> Kpng Sink -----GRPC encoding---> ipvs backend(reassemble diffs back into incremental events) -> program datapath rules  incrementally

Which shows what @mcluseau was talking about above with the reduplication of effort..... 

In a perfect world I think a backend should _try_ to be completely [level driven ](https://hackernoon.com/level-triggering-and-reconciliation-in-kubernetes-1f17fe30333d) which is [recommended by core k8s ](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/controllers.md) HOWEVER this is easier said than done when interacting with APIs (iptables, nft, etc) which are incremental by design. 

At the end of the day if we could change the flow to be more like 

kube api (incremental add/update/delete events) -> Proxystore (keeping fullstate) -> Watchstate (sending explicitly which Services and/or endpoints require a re-sync for a given event) ----GRPC encoding---> ipvs backend (sync COMPLETE state for the relevan service+endpointslices) -> try and program datapath rules in an idempotent manner (again this will look different for various backends) 

_I tried to implement something like this in the ebpf backend_

I think it may be a bit less efficient but: 
- we'll only ever be re-syncing a few services at a time 
- With endpoint slices we will have a cap on the number of endpoints in a given sync cycle
- Being more level driven is much less fragile of a system i.e if for some reason we miss an increment we'll pick up the correct state on the next sync (assuming we're doing idempotent operations in the datapath operations). 

Ultimately IMO we want these backends to be a simple as possible, and for KPNG to provide infrastructure to make that possible.

daman1807:
Thank you @astoycos for the explanation! 
Everything makes much more sense now, I'll definitely go through the documentation/links you shared.

daman1807:
@mcluseau @astoycos @jayunit100 

I have updated the implementation, can you guys review it once?
Have also updated README, and PR Description.

backends/ipvsfullstate/common.go
```diff
@@ -0,0 +1,133 @@
...
+func getTimeoutForClientIPAffinity(service *localv1.Service) int32 {
+	affinity := getSessionAffinity(service.SessionAffinity)
+	if affinity.ClientIP != nil {
+		return affinity.ClientIP.ClientIP.TimeoutSeconds
+	}
+	return 0
+}
+
+// getVirtualServer returns virtual server for the given arguments.
+func getVirtualServer(ip string, service *localv1.Service, portMapping *localv1.PortMapping) *ipvs.VirtualServer {
```
jayunit100, line 75: in general can we replace `getX()` with "CreateX" or "NewX" ?

daman1807: sure, on it.

backends/ipvsfullstate/controller.go
```diff
@@ -0,0 +1,102 @@
...
+func (c *Controller) Callback(ch <-chan *client.ServiceEndpoints) {
+	// TODO do we need a lock here, can we offload the processing to a go routine ?
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	// log time to process the callback
+	st := time.Now()
+	defer func() {
+		et := time.Now()
+		klog.V(3).Infof("took %.2f ms to sync", 1000*et.Sub(st).Seconds())
```
mcluseau, line 76: klog.V(3).Infof("took %v to sync", time.Since(st).Truncate(time.Millisecond)

daman1807:
@jayunit100 @mcluseau I have made the necessary changes, we can merge this now !

jayunit100:
I think we've poked at it enough that we can merge 

Worse case scenario we revert and someone can wrist slap me :)

/lgtm
/approve

PR #426, opened 2023-01-07
Moving command_builder to builder module

knabben:
Moving `command_builder` to a `builder` module in order to be able to use `ToLocalCmd` or `LocalCmds` on external projects.
https://github.com/kubernetes-sigs/windows-service-proxy/pull/1

Adding the backends Register in the main.go instead, so the modules listed on `storecmds` are NOT imported automatically when `builder` is used.


jayunit100:
Can u confirm kpng local up .sh still works after this change b4 merg

jayunit100:
/lgtm
/approve


I'll confirm kpng local up still works tonite

PR #433, opened 2023-01-12
Update tests to use k8s version 1.26.0 plus small change to make test_e2e to work properly again when running parallel kind clusters

qabpea:


qabpea:
A couple of small changes to make the test_e2e script work as it should when running several kind clusters in parallell.
Also updated default k8s version to 1.26.0

qabpea:
Not sure why IPVS fails, it works just fine when I run them

qabpea:
Will recreate the PR, there are other commits to this file that create problems

PR #437, opened 2023-01-25
Added skip filters to make it  possible for ci/cd test succeed

qabpea:
Added skip filters for all combinations of "ip family" and "backend" that are tested when pushing a PR. Also made test_e2e.sh to return 0 at failure and a non zero value at failure


qabpea:
The default right now is to include the skip filters and to use -I to turn them off.
We should add a flag -S to turn them on, then have ci/cd use this flag and then change the default and remove the -I flag

qabpea:
Looks like ginkgo blows up for all the IPVS tests

qabpea:
@astoycos @jayunit100 can you have a look at this PR, test_e2e now has the skip filters in place. All tests except fro ipvs works. I am not sure why all IPVS tests fails, my guess is that it has to do with the linux kernel used by github for the tests.

hack/test_e2e.sh
```diff
@@ -39,6 +39,159 @@ GINKGO_DUMP_LOGS_ON_FAILURE=false
 GINKGO_DISABLE_LOG_DUMP=true
 GINKGO_PROVIDER="local"
 
+# iptables specific skipped ginkgo tests
```
astoycos, line 42: I would recommend putting all these var names in their own file titled like `test_skip_lists.sh` which you could just source, also make each var a single assignment like 

```bash
## Iptables, ipv4
GINKGO_SKIP_ipv4_iptables_TEST="should be updated after adding or deleting ports\n
|should serve multiport endpoints from pods\n
|should check kube-proxy urls\n
etc etc "

GINKGO_SKIP_ipv6_iptables_TEST=...
```

I think that may be more readable?

qabpea: I agree with moving them out into a separate file.
But I find it easier to keep each filter on a separate row.
I have been running the tests over and over and over again to find all filters that must be updated and I found it easier to use row for every filter when adding in the new ones.

This should after all be a temp solution. The next step must be to fix the backends in order to get rid of these skip filters

qabpea: Hmm

What you propose is not \n at the end of each row right, bit a single \
That should be even easier to manage then what is there now

hack/test_e2e.sh
```diff
@@ -41,24 +41,57 @@ GINKGO_PROVIDER="local"
...
+    if [ -z "${result}" ]; then
+        echo "result_message() requires a message"
+        exit 1
+    fi
+    RED="\e[91m"
+    GREEN="\e[92m"    
+    ENDCOLOR="\e[0m"
+
+    if [ "${result}" == "0" ] ; then
+	echo -e "${GREEN}[SUCCES!] ${message} succeded!!! ${ENDCOLOR}"
```
astoycos, line 89:
```suggestion
	echo -e "${GREEN}[SUCCESS!] ${message} succeded!!! ${ENDCOLOR}"
```

qabpea: I managed to fck something up, working on fixing it right now.

qabpea:
Finally, all e2e_tests succeeded for the first time. The test will probably fail now and then anyway. The set of faults for some tests are not stable.

PR #439, opened 2023-01-27
Dummy PR: Explore why test of ipvs do not work right now

qabpea:
This PR will be closed and removed once the problem is understood


qabpea:
It looks like IPVS tests fail when using k8s 1.26.0

qabpea:
OK, the gitlab workflow test bed can not handle ipvs test when k8s version is 1.26.0.

I will cancel this PR and remove the branch

PR #440, opened 2023-01-28
update to k8s version 1.26.0

qabpea:
This will currently fail due to a problem with the github work flow environment. ginkgo will crash for all use of the ipvs backend when executed at github. This does not happen in other environments


mcluseau:
@qabpea rebase and retest, I think. I'm observing a ~50% failure rate on IPVS (flake) so it's quite easy to make it pass.

daman1807:
> @qabpea rebase and retest, I think. I'm observing a ~50% failure rate on IPVS (flake) so it's quite easy to make it pass.

Yeah, github lot of flakes for ipvs on github ci. Everything passes on aws/local machines

qabpea:
I started to update to kind 0.18 and k8s 1.27.1. I see quite a lot of new test failures.
Time to update skip lists again

//Per

qabpea:
I get exactly the same problems as before with a new checkout. I wonder if it for some reason is tied to which location you run git from. I also get the same result with v1.27.1 https://github.com/kubernetes-sigs/kpng/pull/499

The ipvs stuff never get to run the tests, it loops for ever or stops with

PR #443, opened 2023-01-29
IPVS Fullstate Support for IPv6

daman1807:
In continuation to https://github.com/kubernetes-sigs/kpng/pull/417.

Passes e2e for **ipv4,** **ipv6** and **dual** on Ubuntu 22.04.1 LTS.


backends/ipvsfullstate/common.go
```diff
@@ -1,38 +1,73 @@
...
 }
 
-// getNodeIPs safely returns all Node IPs.
-func getNodeIPs() []string {
-	return *NodeAddresses
+// getNodeIPs safely returns all Node IPs for given IPFamily.
+func getNodeIPs(ipFamily v1.IPFamily) []string {
+	IPs := make([]string, 0)
+	for _, ip := range *NodeAddresses {
+		if ipFamily == v1.IPv4Protocol && netutils.IsIPv4String(ip) {
```
jayunit100, line 40: wait cant this just be if IPv4 or IPv6 then
```
IPs = append(...
```

daman1807:
```
func getExternalIPs(service *localv1.Service, ipFamily v1.IPFamily) ([]string, error) {
	if service.IPs.ExternalIPs != nil {
		if ipFamily == v1.IPv4Protocol {
			return service.IPs.ExternalIPs.GetV4(), nil
		} else if ipFamily == v1.IPv6Protocol {
			return service.IPs.ExternalIPs.GetV6(), nil
		}
	}
	return nil, errors.New(fmt.Sprint("ipfamily - not supported", ipFamily))
}
```
@jayunit100 something like this ?

daman1807: > wait can't this just be IPv4 or IPv6 then

Entries in an IPSet can only belong to one IPFamily, so I loop service creation logic for both the ipFamilies.
```
// iterate over ipFamily
for _, ipFamily := range []v1.IPFamily{v1.IPv4Protocol, v1.IPv6Protocol} {
    / /iterate over NodeIPs
    for _, nodeIP := range getNodeIPs(ipFamily) {
        ....
    }
}
```
And then I pick the relevant IPSet accordingly 
```
kubeNodePortTCPIPSet  = map[v1.IPFamily]string{
    v1.IPv4Protocol: "KUBE-NODE-PORT-TCP",
    v1.IPv6Protocol: "KUBE-6-NODE-PORT-TCP",
}
```

daman1807:
@jayunit100 all the tests (Security / govulncheck) pass now, we can merge this.

PR #460, opened 2023-02-09
misc service-proxy.md updates

danwinship:
- link to end-user service/proxy docs
- fix reference to userspace proxy, which no longer exists upstream
- rewrite some stuff where I was trying to pluralize field names in a way that looks bad in the rendered doc
- update recommendations around localhost nodeports ("don't!")
- assume ProxyTerminatingEndpoints (since it's expected to be GA in 1.27), and describe it as applying to all services, not just eTP:Local ones (cf https://github.com/kubernetes/kubernetes/pull/108691)
- fix definition of "internal" vs "external" connections to be destination-based, not source-based (cf https://github.com/kubernetes/kubernetes/issues/108526)
- explain _why_ eTP:Local drops bad traffic rather than rejecting it
- rewrite "Port Listeners" section to describe it as something we _used to_ do rather than something we do (cf https://github.com/kubernetes/kubernetes/pull/108496)

/cc @aojea @tssurya @astoycos

### DIFF
--- a/doc/service-proxy.md
+++ b/doc/service-proxy.md
@@ -13,7 +13,14 @@ network plugins to implement NetworkPolicy correctly without making
 assumptions about the behavior of the service proxy.
 
 (The rest of this document assumes that you already generally
-understand Kubernetes Services from an end-user point of view.)
+understand [Kubernetes
+Services](https://kubernetes.io/docs/concepts/services-networking/)
+and [the service
+proxy](https://kubernetes.io/docs/reference/networking/virtual-ips/)
+from an end-user point of view. In particular, it does not bother to
+explain certain service features whose behavior on the service proxy
+side is already fully understandable just from the end-user
+documentation.)
 
 ## Note on Dual Stack, and Legacy vs "Modern" APIs
 
@@ -45,12 +52,13 @@ proxying by DNAT'ing (and sometimes SNAT'ing) connections. Any service
 proxy implementation must do something at least vaguely equivalent to
 this, but it may not always literally be NAT.
 
-For example, the userspace proxy (which, admittedly, is deprecated),
-does not NAT connections, but instead accepts inbound connections to
-service IPs in userspace, and then creates new outbound connections to
-the endpoint IPs, and then copies data between the connections. (Thus,
-all connections passing through the userspace proxy will arrive at
-their endpoint with a node IP as their source IP.)
+For example, kpng's userspace proxy (and the now-removed upstream
+userspace proxy it was based on), does not NAT connections, but
+instead accepts inbound connections to service IPs in userspace, and
+then creates new outbound connections to the endpoint IPs, and then
+copies data between the connections. (Thus, all connections passing
+through the userspace proxy will arrive at their endpoint with a node
+IP as their source IP.)
 
 In other cases, some proxies may be able to do clever hacks to avoid
 NAT'ing (especially SNAT'ing) that would otherwise be needed.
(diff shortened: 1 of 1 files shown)


doc/service-proxy.md
```diff
@@ -105,11 +125,12 @@ to ensure that it does. Eg:
...
     network namespace for un-DNAT'ing first.
 
-  - Likewise, if a host-network process tries to connect to a NodePort
-    service via `127.0.0.1` or `::1`, it is necessary to SNAT the
-    packet, since otherwise the destination pod would think the packet
-    came from its own `127.0.0.1`/`::1`. (In fact, the kernel will log
-    a warning if you try to DNAT-but-not-SNAT a packet whose source IP
+  - Likewise, if the proxy supports localhost NodePort connections,
+    and a host-network process tries to connect to a NodePort service
+    via `127.0.0.1` or `::1`, it is necessary to SNAT the packet,
```
aojea, line 130: this is the route_localnet hack, no? 

this doesn't work for IPv6 no matter what https://github.com/kubernetes/kubernetes/issues/90236

doc/service-proxy.md
```diff
@@ -219,18 +240,30 @@ circuited by the service proxy.
...
+cluster. If there are no local endpoint IPs on that node then external
+connections to the service via that node should be dropped.
+
+(For external traffic policy in particular, it is important that
+"wrongly-directed" traffic be _dropped_, not _rejected_, because there
+is a race condition between when the proxy stops accepting new
+connections for a service on a particular node versus when the load
+balancer _notices_ that the proxy has stopped accepting new
+connections. During that time, if the load balancer sends a connection
+to a "bad" node, we drop the packets so that the client will think
```
aojea, line 262: the client will eventually timeout and receive an error, since the tcp stack retries with the same tuple is not likely the  intermediate devices will reroute hte packets to other node

```
tcpdump -i any -n tcp port 80
tcpdump: data link type LINUX_SLL2
tcpdump: verbose output suppressed, use -v[v]... for full protocol decode
listening on any, link-type LINUX_SLL2 (Linux cooked v2), snapshot length 262144 bytes
14:28:07.117111 eth0  Out IP 192.168.8.2.58830 > 10.96.61.91.80: Flags [S], seq 3360558320, win 64240, options [mss 1460,sackOK,TS val 2346934370 ecr 0,nop,wscale 7], length 0
14:28:08.119044 eth0  Out IP 192.168.8.2.58830 > 10.96.61.91.80: Flags [S], seq 3360558320, win 64240, options [mss 1460,sackOK,TS val 2346935372 ecr 0,nop,wscale 7], length 0
14:28:10.135037 eth0  Out IP 192.168.8.2.58830 > 10.96.61.91.80: Flags [S], seq 3360558320, win 64240, options [mss 1460,sackOK,TS val 2346937388 ecr 0,nop,wscale 7], length 0
14:28:14.203047 eth0  Out IP 192.168.8.2.58830 > 10.96.61.91.80: Flags [S], seq 3360558320, win 64240, options [mss 1460,sackOK,TS val 2346941456 ecr 0,nop,wscale 7], length 0
```


the reject will indicate something is wrong and the client will be able to retry sooner, in both cases AFAIK the client will have a network error

doc/service-proxy.md
danwinship: Ah! I noticed that this contradicted [the comments in the kube-proxy source code](https://github.com/kubernetes/kubernetes/blob/8f94681cd294aa8cfd3407b8191f6c70214973a4/pkg/proxy/iptables/proxier.go#L1437-L1439) but I was thinking this doc was right and the code was wrong.

doc/service-proxy.md
```diff
@@ -59,30 +67,42 @@ NAT'ing (especially SNAT'ing) that would otherwise be needed.
...
+        connections on `127.0.0.1` (but not `::1`), though the ipvs
+        backend does not, and this behavior is considered deprecated
+        and can be disabled.
+
+If a service has at least one usable endpoint, then connections to any
+of the above destinations should be randomly DNAT'ed to one of the
+usable endpoint IPs (with the destination port changed to the
+corresponding `.spec.port[].targetPort` value, if that is set). If a
+service does not have any usable endpoints, then connections to any of
+the above destinations should be rejected (ie, actively refused with
```
astoycos, line 96: Should the ICMP error always be a [Type 3 - Destination Unreachable ](https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml#icmp-parameters-codes-3) packet? Is that specified somewhere?

aojea: I think that is what iptables does by default with `-j REJECT` and we carrying over with that , maybe we should indicate that it should reject the packet and leave that open, you can also use TCP_RST

astoycos: Right for UDP I think we'd have to use ICMP but for TCP that makes sense

doc/service-proxy.md
danwinship: The dropping behavior has been intentional since the beginning (["KEP"](https://github.com/kubernetes/community/blob/f4379b3c775e31b38df46fb174cdafceceaaca33/contributors/design-proposals/external-lb-source-ip-preservation.md#external-traffic-blackholed-on-nodes-with-no-local-endpoints), [implementation](https://github.com/kubernetes/kubernetes/pull/29409/files#diff-2ccafd63d57233fd0697b9852514604ab880e29aeb073035765ad7541cad1b35R1168)). It's possible that we've been doing it wrong all along, I guess?

doc/service-proxy.md
danwinship: no, but... none of the other ICMP types really imply "reject" so it's strongly implied.

(Hm... and looking at the [ICMP code to errno conversion table](https://github.com/torvalds/linux/blob/v6.1/net/ipv4/icmp.c#L113) it looks like if we returned `ICMP_PORT_UNREACH` instead of the default `ICMP_HOST_UNREACH` then the caller would get back "Connection refused" instead of "No route to host"... Maybe we should start doing that, and recommend that...)

aojea: ICMP_PORT_UNREACH is correct, you connect to a service IP:port and the port is not reachable because doesn't have endpoints

aojea: is the iptables default for `-j REJECT` , isn't it?

PR #480, opened 2023-04-07
expose metrics endpoint from within kpng

mattfenwick:
# What

Sets up prometheus server in kpng that can be scraped by prometheus master.

Sets up Grafana with prometheus as a data source.

Sets up a couple of default dashboards (for kube api server and nginx ingress controller).

Sets up an nginx ingress controller , so that Grafana can be accessed through ingress.

For some reason, the existing `--exposeMetrics` flag DOES NOT work => there seem to be problems with flag setup in the kpng cli.  Therefore, this adds a new function to set up the metrics server.

# How to use

See new hack/metrics/README.md

### DIFF
--- a/cmd/kpng/main.go
+++ b/cmd/kpng/main.go
@@ -44,10 +44,16 @@ var (
 
 // main starts the kpng program by running the command sent by the user.  This is the entry point to kpng!
 func main() {
+	metricsPort := "9099"
+	setupPrometheusServer(context.TODO(), fmt.Sprintf("0.0.0.0:%s", metricsPort))
+
 	klog.InitFlags(flag.CommandLine)
 
 	cmd := cobra.Command{
 		Use: "kpng",
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			klog.Infof("persistent pre run: %s\n", cmd.Name())
+		},
 	}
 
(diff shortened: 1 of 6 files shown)


cmd/kpng/main.go
```diff
@@ -44,10 +44,16 @@ var (
 
 // main starts the kpng program by running the command sent by the user.  This is the entry point to kpng!
 func main() {
+	metricsPort := "9099"
```
jayunit100, line 47: can u add a log statement in a follow on pr at some point saying metrics are on that port

jayunit100 (approved):

cmd/kpng/main.go
mattfenwick: ` I0407 00:32:48.665494       1 metrics.go:38] Starting metrics server at 0.0.0.0:9099  ` already shows up in the logs, is that good enough?

mattfenwick: In context:

```
kubectl logs -f -n kube-system kpng-9dzjs kpng
I0407 00:32:48.665334       1 main.go:78] exporting metrics to: 0.0.0.0:9099 
I0407 00:32:48.665494       1 metrics.go:38] Starting metrics server at 0.0.0.0:9099
...
```

PR #501, opened 2023-04-27
Update e2e_tests.yaml

qabpea:
Removed ipvs testing for now.
We need to figure out how to get ipvs work again.
It seems we need to load the driver and do some other setup for ipvs to work now

Fixes #


qabpea:
Thanks, I will trigger all the test again for the k8s 1.27.1 integration and cancel the integration of 1.26.3

//Per

PR #505, opened 2023-05-02
rename EndpointsClient to LocalClient

### LINKED ISSUE #484: comprehensive comments/docs on the Endpoints client and the kpng/client/client.go
The *EndpointsClient* is a little confusing to read.  

- I THINK It doesnt specifically filter on endpoints, and in any case, we have endpointslices, services, and endpoints, which i believe it filters and watches on.  

- Also it doesnt say much about how the watch mechanism works and how its node specific

```
// doesnt EndpointsClient watch everything ? it sais its just a client to the endpoints API...
// EndpointsClient is a simple client to kube-proxy's Endpoints API.
type EndpointsClient struct {
	// Target is the gRPC dial target
	Target string
```

abhaikollara:
### What kind of PR is this?
Improvement

See https://github.com/kubernetes-sigs/kpng/issues/484#issuecomment-1522807182. The PR renames the `EndpointsClient` to `LocalClient` as per the suggested comment

Fixes #484

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -41,19 +41,19 @@ type ServiceEndpoints = fullstate.ServiceEndpoints
 
 type FlagSet = tlsflags.FlagSet
 
-// New returns a new EndpointsClient with values bound to the given flag-set for command-line tools.
-// Other needs can use `&EndpointsClient{...}` directly.
-func New(flags FlagSet) (epc *EndpointsClient) {
-	epc = &EndpointsClient{
+// New returns a new LocalClient with values bound to the given flag-set for command-line tools.
+// Other needs can use `&LocalClient{...}` directly.
+func New(flags FlagSet) (lc *LocalClient) {
+	lc = &LocalClient{
 		TLS: &tlsflags.Flags{},
 	}
-	epc.ctx, epc.cancel = context.WithCancel(context.Background())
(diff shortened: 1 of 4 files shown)


client/client.go
```diff
@@ -26,9 +26,6 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 
-	// allow multi gRPC URLs
-	//_ "github.com/Jille/grpc-multi-resolver"
```
daman1807, line 30: are you sure we don't need any initialization to happen here?

abhaikollara: The code was already commented out, was just cleaning it up. 

I'm guessing, even if the init is required, it should be part of another PR ?

daman1807: oh, let it be as it is then :)

abhaikollara: Have reverted my change

daman1807:
@abhaikollara can you squash the two commits

abhaikollara:
> @abhaikollara can you squash the two commits

Done

daman1807:
/lgtm
thanks !

daman1807:
deferring to @knabben for final approval, I don't have much context here.

mcluseau:
that will make @jayunit100  happy ;)
/approve

PR #520, opened 2023-08-14
Populate port mapping name to make endpoint port mapping work.

mneverov:
### What kind of PR is this?
Bug fix: previously, port name was not populated that prevented endpoint port mapping to work correctly.

This PR fix the issue above.

Fixes #
Fixes tests:
`should be updated after adding or deleting ports` and `should serve multiport endpoints from pods  [Conformance]` in `e2e-ipv4-iptables` mode.


mneverov:
question: should I remove `should be updated after adding or deleting ports` and `should serve multiport endpoints from pods  [Conformance]` tests from [ginkgo skip lists for ipv4 and ipv6](https://github.com/kubernetes-sigs/kpng/blob/master/hack/test_skip_list.sh#L20)?

PR #523, opened 2023-08-25
Fix artifacts directory path.

mneverov:
### What kind of PR is this?
Bugfix.

Previously, reports were stored under `temp/e2e/bin/artifacts/reports` [see ginkgo command](https://github.com/kubernetes-sigs/kpng/blob/master/hack/test_e2e.sh#L582), so the `upload junit reports` step [was not working](https://github.com/kubernetes-sigs/kpng/actions/runs/5942514073/job/16135449056#step:7:11).
This PR aligns artifacts directory so it is used for logs and reports.

Fixes #


jayunit100:
thanks you sir !
/lgtm
/approve

Issue #467, opened 2023-03-18
kpng local up should default to iptables: it fails w/ nft


jayunit100:
I just noticed hack/kpng-local-up.sh fails w/ NFT mode but works w/ iptables. 
i wonder why nft wouldnt be able to write the `kubernetes` rules by default ? 

```
[kpng-sbffw kpng-nft]   jump z_filter_all
[kpng-4tsfl kpng-nft]  }
[kpng-sbffw kpng-nft]  }
[kpng-cbb2k kpng-nft]  chain z_hook_filter_output {
[kpng-sbffw kpng-nft]  chain z_hook_filter_output {
[kpng-4tsfl kpng-nft]  chain z_hook_filter_output {
[kpng-cbb2k kpng-nft]   type filter hook output priority 0;
[kpng-4tsfl kpng-nft]   type filter hook output priority 0;
[kpng-sbffw kpng-nft]   type filter hook output priority 0;
[kpng-cbb2k kpng-nft]   jump z_filter_all
[kpng-cbb2k kpng-nft]  }
[kpng-4tsfl kpng-nft]   jump z_filter_all
[kpng-sbffw kpng-nft]   jump z_filter_all
[kpng-cbb2k kpng-nft] }
[kpng-4tsfl kpng-nft]  }
[kpng-sbffw kpng-nft]  }
[kpng-4tsfl kpng-nft] }
[kpng-cbb2k kpng-nft] /dev/stdin:11:3-21: Error: Could not process rule: No such file or directory
[kpng-sbffw kpng-nft] }
[kpng-cbb2k kpng-nft]   numgen random mod 1 vmap {
[kpng-sbffw kpng-nft] /dev/stdin:11:3-21: Error: Could not process rule: No such file or directory
[kpng-4tsfl kpng-nft] /dev/stdin:11:3-21: Error: Could not process rule: No such file or directory
[kpng-cbb2k kpng-nft]   ^^^^^^^^^^^^^^^^^^^
[kpng-sbffw kpng-nft]   numgen random mod 1 vmap {
[kpng-4tsfl kpng-nft]   numgen random mod 1 vmap {
[kpng-sbffw kpng-nft]   ^^^^^^^^^^^^^^^^^^^
[kpng-cbb2k kpng-nft] /dev/stdin:67:3-16: Error: Could not process rule: No such file or directory
[kpng-4tsfl kpng-nft]   ^^^^^^^^^^^^^^^^^^^
[...]

PR #525, opened 2023-08-25
Downgrade kpng alpine image to align nftables with nodes OS.

mneverov:
### What kind of PR is this?

Hopefully a bugfix.

This PR aligns `nftables` version in [kpng image](https://pkgs.alpinelinux.org/packages?name=nftables&branch=v3.16&repo=&arch=x86_64&maintainer=) and [os node image](https://packages.ubuntu.com/jammy/nftables).
Currently, when e2e tests run locally nft rules set up incorrectly due to the versions mismatch that results in `[invalid type]` in chains.
To reproduce 
1. start `kpng` without running tests via `./hack/test_e2e.sh -i ipv4 -b nft -d`
2. join a worker node, install `nftables` and check for `invalid type`:
```sh
kubectl node-shell kpng-e2e-ipv4-nft-worker
apt update && apt install -y nftables
nft list ruleset | grep invalid
```

`kpng` has nothing to do with that. One can reproduce it by ssh-ing to a kpng pod and applying nft rules:
```sh
vi nft.conf
table ip6 test {
        chain svc_1 {
                tcp dport 80 dnat to fd6d:706e:6701:2::2:8080
                fib daddr type local tcp dport 30202 dnat to fd6d:706e:6701:2::2:8080
        }

        chain svc_default_nodeport-test_eps {
                numgen random mod 2 vmap { 0 : jump svc_2, 1 : jump svc_1 }
        }

        chain svc_2 {
                tcp dport 80 dnat to fd6d:706e:6701:1::2:8080
                fib daddr type local tcp dport 30202 dnat to fd6d:706e:6701:1::2:8080
        }
}
nft -f nft.conf
```

On the node the new `test` table will look like:
```
table ip6 test {
        chain svc_1 {
                tcp dport 80 dnat to fd6d:706e:6701:2::2:8080
                fib daddr type local tcp dport 30202 dnat to fd6d:706e:6701:2::2:8080
        }

        chain svc_default_nodeport-test_eps {
                numgen random mod 2 map { 0 : 0x0 [invalid type], 1 : 0x1000000 [invalid type] }
        }

        chain svc_2 {
                tcp dport 80 dnat to fd6d:706e:6701:1::2:8080
                fib daddr type local tcp dport 30202 dnat to fd6d:706e:6701:1::2:8080
        }
}
```

There is no `invalid type` message when the same steps are executed from `alpine:3.16` image.

**NOTE**: I checked GH e2e action and it uses the same images (alpine 3.17 and ubuntu 22.04). When tests run locally I see invalid types during tests execution, but 335 tests run successfully so I suspect the chains with invalid types are not hit :shrug:.

Probably related to the [issue](https://github.com/kubernetes-sigs/kpng/issues/467). 

Fixes #

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ add . ./
 run go install -trimpath -buildvcs=false ./cmd/...
 
 # the real image
-from alpine:3.17
+from alpine:3.16
 entrypoint ["/bin/kpng"]
 run apk add --update iptables ip6tables iproute2 ipvsadm nftables ipset conntrack-tools
 copy --from=build /go/bin/ /bin/


jayunit100:
ok so this was the reason why it stopped working recently in the hack/local-up-kpng recipe i guesS?

mneverov:
> ok so this was the reason why it stopped working recently in the hack/local-up-kpng recipe i guesS?

it might be related, but also the error you had is different...

mcluseau:
alpine is 3.18 now, I'm not sure we should downgrade for a specific node version. nftables is an in-kernel VM and nft compiles the given rules for this VM, hence it works in the kernel, even if the "decompilation" with an older version of nft is not guaranteed to work 100%. Downgrading nft means to going back in time, so less bugs fixed. To get the expected result from commands like `nft list ruleset`, it's better to use the nft instance in the kpng pod, not the one on the node.

mneverov:
> alpine is 3.18 now, I'm not sure we should downgrade for a specific node version. nftables is an in-kernel VM and nft compiles the given rules for this VM, hence it works in the kernel, even if the "decompilation" with an older version of nft is not guaranteed to work 100%. Downgrading nft means to going back in time, so less bugs fixed. To get the expected result from commands like `nft list ruleset`, it's better to use the nft instance in the kpng pod, not the one on the node.

yes, you're right. It explains why tests ran successfully on alpine 3.17. Sorry for the noise, ptal #527. Please let me know if I can bump golang build image to the latest alpine too and golang version to 1.21.

mcluseau:
@mneverov hey, no wrrie, thanks for contirbuting, there's no building without some noise :-)

PR #526, opened 2023-08-27
docker: upgrade to golang 1.21 and alpine 3.18

mcluseau:
Upgrade versions used in the docker, and also upgrade Ginkgo so it's the same as the CLI, allowing the tests to run.

PR #527, opened 2023-08-27
Bump kpng alpine image.

mneverov:
### What kind of PR is this?
This PR reverts #525 and also bumps alpine to 3.18

Decompilation of the nft ruleset with the older nftables version might not work, but the rules themselves should work in kernel correctly. See [comment](https://github.com/kubernetes-sigs/kpng/pull/525#issuecomment-1694610205).

Fixes #


mcluseau:
I think it's a dup of #526 :-)

PR #532, opened 2023-09-23
Fix epslices

mneverov:
### What kind of PR is this?

Bugfix

First commit addresses a situation when an endpoint slice is not ready, but it was not set.
Second commit complements #520 (which can be closed in favor of this PR). The [port mapping](https://github.com/kubernetes-sigs/kpng/commit/6ec42e7b6ce414092d406ea1c9bc75bd849cea75#diff-797a7515d4f869cd70ff1e72a62fab4fd348ad3b509f9a815dbe45ffcbcb3638R34) should be done based on field precedence: 1. `port.Name`; 2. `port.TargetPortName`; 3. `port.TargetPort` (see the [adjusted test](https://github.com/kubernetes-sigs/kpng/commit/6ec42e7b6ce414092d406ea1c9bc75bd849cea75#diff-6979f344dc49b8ed56f05d876a188b024bbf09e14ebb5131323950208bae9a8fR54)).

Fixes #
Following tests pass:
- `EndpointSlice [It] should support a Service with multiple ports specified in multiple EndpointSlices`
- `EndpointSlice [It] should support a Service with multiple endpoint IPs specified in multiple EndpointSlices`
- `EndpointSliceMirroring [It] should mirror a custom Endpoint with multiple subsets and same IP address`

### DIFF
--- a/backends/iptables/iptables.go
+++ b/backends/iptables/iptables.go
@@ -647,10 +647,15 @@ func (t *iptables) createEndpointsChain(svcInfo *serviceInfo, allEndpoints *endp
 			ep = epInfo.IPs.V4[0]
 		}
 
-		targetPort := epInfo.PortMapping(&localv1.PortMapping{
+		targetPort, err := epInfo.PortMapping(&localv1.PortMapping{
+			Name:           svcInfo.portName,
 			TargetPortName: svcInfo.targetPortName,
 			TargetPort:     int32(svcInfo.targetPort),
 		})
+		if err != nil {
+			klog.V(1).InfoS("failed to map port", "err", err)
+			continue
+		}
 		endpointPortMap[ep] = targetPort
 		endpoints = append(endpoints, &ep)
 
@@ -748,21 +753,15 @@ func (t *iptables) writeDNATRules(svcInfo *serviceInfo, svcName types.Namespaced
 
 		targetPort := t.getTargetPort(svcInfo, endpointPortMap, *epIP)
 
-		// this seems very sly to me. Doing this because there were 2 entries being added
-		// one with the right target port and one with zero
-		// write better logic or verify how baseServiceInfo & endpointPortMap are populated
-		if targetPort == 0 {
-			continue
-		}
 		// DNAT to final destination.
 		args = append(args, "-m", protocol, "-p", protocol, "-j", "DNAT", "--to-destination", net.JoinHostPort(*epIP, strconv.Itoa(targetPort)))
 		t.natRules.Write(args)
 	}
 }
 
-// if the targetPort is string, fetch the value from endpointPortMap
+// if the targetPort is string or portName exists, fetch the value from endpointPortMap
 func (t *iptables) getTargetPort(svcInfo *serviceInfo, endpointPortMap map[string]int32, endpoint string) int {
-	if svcInfo.TargetPortName() != "" {
+	if svcInfo.PortName() != "" || svcInfo.TargetPortName() != "" {
 		return int(endpointPortMap[endpoint])
 	}
 	return svcInfo.TargetPort()
--- a/api/go.mod
+++ b/api/go.mod
@@ -6,9 +6,11 @@ require (
 	github.com/golang/protobuf v1.5.3
 	google.golang.org/grpc v1.50.0
 	google.golang.org/protobuf v1.28.1
+	k8s.io/klog/v2 v2.80.1
 )
 
 require (
+	github.com/go-logr/logr v1.2.4 // indirect
 	github.com/google/go-cmp v0.5.9 // indirect
 	golang.org/x/net v0.14.0 // indirect
 	golang.org/x/sys v0.11.0 // indirect
--- a/api/go.sum
+++ b/api/go.sum
@@ -1,3 +1,4 @@
+github.com/go-logr/logr v1.2.4 h1:g01GSCwiDw2xSZfjJ2/T9M+S6pFdcNtFYsp+Y43HYDQ=
 github.com/golang/protobuf v1.5.0/go.mod h1:FsONVRAS9T7sI+LIUmWTfcYkHO4aIWwzhcaSAoJOfIk=
 github.com/golang/protobuf v1.5.3 h1:KhyjKVUg7Usr/dYsdSqoFveMYd5ko72D+zANwlG1mmg=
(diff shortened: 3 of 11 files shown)


jayunit100:
i think we can merge this now.  it looks good to me.
we can test/look at it together tomorrow
/lgtm
/approve

backends/iptables/iptables.go
```diff
@@ -748,21 +753,15 @@ func (t *iptables) writeDNATRules(svcInfo *serviceInfo, svcName types.Namespaced
...
 		args = append(args, "-m", protocol, "-p", protocol, "-j", "DNAT", "--to-destination", net.JoinHostPort(*epIP, strconv.Itoa(targetPort)))
 		t.natRules.Write(args)
 	}
 }
 
-// if the targetPort is string, fetch the value from endpointPortMap
+// if the targetPort is string or portName exists, fetch the value from endpointPortMap
 func (t *iptables) getTargetPort(svcInfo *serviceInfo, endpointPortMap map[string]int32, endpoint string) int {
-	if svcInfo.TargetPortName() != "" {
+	if svcInfo.PortName() != "" || svcInfo.TargetPortName() != "" {
```
jayunit100, line 764: i wonder if in general , KPNG NEEDS to be dependent on port names for routing rules.  To me it shoudlnt. 

i.e. if

- service 10.1.2.3:8080 points to my hamburger SAAS
- service 10.1.2.3:9090 points to my pizza SAAS

Why cant we blindly just write routing rules to 8080 and 9090.  Who cares what the names are.  After all, the `10.1.2.3:8080` and `10.1.2.3:9090` represent unique primary keys, and are complete in terms of the layer 4 information content required to distinguish these backends.? right?

jayunit100: Ah got it.
Kubernetes supports this odd thing where `targetPort=80` and portName=port8080 ... In that case , 
- port name overrides targetPort
but KPNG discarded that name info and then forwareded to the incorrect,  targetPort 80  which was supposed to be replaced w 8080

mcluseau: umm isn't what we provide with port overrides in endpoints? I mean, the PortMapping(...) call in api/localv1/endpoint.go

mneverov: @mcluseau previously, the [PortMapping](https://github.com/kubernetes-sigs/kpng/pull/532/files#diff-797a7515d4f869cd70ff1e72a62fab4fd348ad3b509f9a815dbe45ffcbcb3638L34) only worked if `TargetPortName` was specified. This PR added port name check too: `if port.Name != "" ...`.

jayunit100: @mikael maybe suggesting there a possible brain fix here ???

mcluseau: well, the idea is still to keep k8s business logic outside of the backends. Otherwise, we're just doing that https://xkcd.com/927/ ;-)
As I'm not very available on the project nowadays, I couldn't dig this a lot, but I can see that @mneverov actually did some work in the file I mentioned, so he can have missed it :-) Now, I just wonder why there's a second round of interpretation has, from my "old" memories, the idea of the `PortMappings()` calls was to directly provide the "final form" of argument `enpointPortMap` argument here in `iptables.go`.

mneverov: > `well, the idea is still to keep k8s business logic outside of the backends`

I was not aware of it, got it.

> `the idea of the PortMappings() calls was to directly provide the "final form" of argument enpointPortMap argument here in iptables.go`

From my perspective this PR did not change it.

My interpretation of how port mapping works:

### K8s
Endpoint slice [reconciler](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/endpointslice/reconciler.go#L161) creates endoint slices with port names taken from a corresponding [service port name](https://github.com/kubernetes/kubernetes/blob/fe84992651335ba6dd66e16754e8e416e5144b59/staging/src/k8s.io/endpointslice/utils.go#L87). A Service port [binds](https://github.com/kubernetes/kubernetes/blob/fe84992651335ba6dd66e16754e8e416e5144b59/staging/src/k8s.io/endpointslice/utils.go#L390) `targetName` to the Pod port if the `targetName` is a string (as it is described in [the documentation](https://kubernetes.io/docs/concepts/services-networking/service/#field-spec-ports)). The same [logic](https://github.com/kubernetes/kubernetes/blob/fe84992651335ba6dd66e16754e8e416e5144b59/pkg/api/v1/pod/util.go#L32) is used in the [endpoints controller](https://github.com/kubernetes/kubernetes/blob/fe84992651335ba6dd66e16754e8e416e5144b59/pkg/controller/endpoint/endpoints_controller.go#L443). If a Service port `targetName` is specified as a string and it does not match to any Pod port names then the controllers create an EndpointSlice (or legacy Endpoints) without this port. In case if it is the only port specified in a Service then EndpoinSlice will not have ports at all. If the `targetName` is an integer (i.e. a port 8080), then it is copied to the EndpointSlice port definition.

### KPNG 
KPNGs frontend kube2store watches api-server and [maps](https://github.com/kubernetes-sigs/kpng/blob/6ccd640378e22cba27f119f7ae132026722a5bfc/server/jobs/kube2store/service-event-handler.go#L113) service ports to the internal PortMapping representation. The only difference from k8s is that PortMapping has a separate field `TargetPortName` for the case when the service port is a string.

So, actually [this check](https://github.com/kubernetes-sigs/kpng/pull/532/files#diff-797a7515d4f869cd70ff1e72a62fab4fd348ad3b509f9a815dbe45ffcbcb3638R38) doesn't make sense and should be removed.

For services without selectors the mapping rules are almost the same:
1. If a Service specifies a port name, then it has to match with the EndpointSlice port name. If it doesn't - no routing. One corner case is when Service doesn't specify port name, but EndpointSlice does. In this scenario again, no routing.
2. If a Service specifies `targetPort` (doesn't matter if it is as a string or actual port) it is not part of the decision, i.e. it can be any bogus port and as long as there is a properly defined EndpointSlice, the routing will be created.

```yaml
apiVersion: v1
kind: Service
metadata:
  name: nginx-service
spec:
  ports:
    - port: 80
      protocol: TCP
      targetPort: anything-here

---

apiVersion: discovery.k8s.io/v1
kind: EndpointSlice
metadata:
  name: nginx-service-1
  labels:
    kubernetes.io/service-name: nginx-service
addressType: IPv4
ports:
  - name: ''
    protocol: TCP
    port: 80
endpoints:
  - addresses:
      - "10.244.0.5" # ip of the nginx port
```

